PR #1, opened 2016-09-13
ENH: initial plugin structure and actions

gregcaporaso:
@benjjneb, this creates the ``q2-dada`` plugin and adds support for your two DADA2 R scripts. The [README.md](https://github.com/gregcaporaso/q2-dada2/blob/b7929ee0f01b4c60829c5f1a3d6e5bb1245010dd/README.md) has been updated to illustrate how to use this (including installation of development versions of QIIME 2 that are necessary to get this to work). 

This is ready for review/merge. I've tested the instructions in the README on my end, but it would be great if you could test as well (see my note below about the bioconda install on Linux though, if you're working on Linux). 

A couple of questions:
- Would it be possible for you to either add a filtering/trimming step to the existing ``run_dada.R`` script, or for you to add a new R script that would do that (#3)? I thought we had discussed that being a part of it, but maybe not. I ultimately want this to be able to use this plugin starting from data like you use in your tutorial (i.e., demultiplexed fastq, with no other processing having yet been applied).
- Are there other options for feature (i.e., OTU) naming in the resulting table than the full sequences? I can work with that, but I need to have this generate a fasta file of the sequences (#2), and it'll be larger than it needs to be if the sequences are duplicated. If they were named for example with the md5 sum of the sequence that would help. If this isn't an existing option in DADA2, don't worry about it. 
- Does the citation, website and help information presented here what you want? I can make any necessary changes. 

```bash
$ qiime dada2 --help
Usage: qiime dada2 [OPTIONS] COMMAND [ARGS]...

  Plugin website: http://benjjneb.github.io/dada2/

  Getting user support: To get help with DADA2, post to the DADA2 issue
  tracker: https://github.com/benjjneb/dada2/issues

  Citing this plugin: DADA2: High-resolution sample inference from Illumina
  amplicon data. Benjamin J Callahan, Paul J McMurdie, Michael J Rosen,
  Andrew W Han, Amy Jo A Johnson, Susan P Holmes Nature Methods 13, 581–583
  (2016) doi:10.1038/nmeth.3869.
```

Outstanding issues:
- I think the bioconda install of DADA2 doesn't work on Linux, so the readme instructions will only work on Mac at the moment. I need to look into this more.


benjjneb:
Looks good at first glance, but will test before merging, and think about filtering. Since just forward reads for now a pre-defined set of filter/trim parameters will work pretty well, but what is the longer-term mechanism for allowing the user to pass in parameters?

Changing names in the table is no problem.

We just headed to Hawaii for a week vacation so I might be a bit slow til I get back.

gregcaporaso:
Thanks @benjjneb!

> Since just forward reads for now a pre-defined set of filter/trim parameters will work pretty well

Do you have a recommendation for what those defaults could be? 

>  but what is the longer-term mechanism for allowing the user to pass in parameters?

There are no limitations with users passing parameters right now. For example, in the ``plot-qualities`` visualizer that I wrapped around your R script, the user passes ``n``, which is the number of ``fastq.gz`` files to generate plots for (``n`` of the ``fastq.gz`` files are selected at random). You might be confusing this with use in Qiita, where the developers want to define a set of default parameters that can be use across runs? 

I'll likely continue working on this in the next week while you're on vacation - I might try to add the filtering/trimming to the R script myself as I'm anxious to try it out on test data. I'll do all of this work on my master branch and have it update this pull request as I go.

Issue #33, opened 2017-01-14
test latest dada2 bioconda package when available


jairideout:
Currently we are using Bioconductor to install the latest version of dada2 (1.2.1). There is work being done to package this release via bioconda; see https://github.com/qiime2/q2-dada2/issues/13 for previous discussion. When the package is available we'll want to test it with q2-dada2 and update the qiime2 install docs.

cc: @zachcp

benjjneb:
Can this be closed? Seems like bioconda install is now working.

ebolyen:
Good catch!

Issue #31, opened 2016-12-27
Learn error rates from a subset of the data


benjjneb:
Parameter learning is the most computationally part of the "tutorial" workflow currently implemented. It is much faster, and achieves essentially the same result, to learn the error rates from a subset of the data for larger datasets. This can substantially speed computation times for large datasets (2x-7x faster).

Workflow should add an argument, perhaps `--nreads-learn` with a default of perhaps 1-million, specifying the minimum number of reads used to learn the error rates. Those error rates will then be used to process the entire dataset.

This approach follows that in the DADA2 "big data" workflow: http://benjjneb.github.io/dada2/bigdata.html

gregcaporaso:
This sounds like it would be really helpful. Think we could get this in with the for the next release?

benjjneb:
Yes.

benjjneb:
Added in 685ddae848e46090aea61809373e7c86757f0676 and b66432da308a924a26f315a517c934b4a335af2f

Assumes dada2 1.2+

Issue #16, opened 2016-10-10
better error message when truncating past max seq length


jairideout:
When passing `--p-trunc-len` with a length greater than the length of the input sequences, `dada2 denoise` (which runs `run_data.R`) fails with an uninformative error message. For example, running this command (`--p-trunc-len 98`) on sequences that have 90 positions:

``` shell
$ qiime dada2 denoise --i-demultiplexed-seqs 88soils-tutorial-demux-1p.qza --o-representative-sequences rep-seqs --o-table table --p-trim-left 0 --p-trunc-len 98 --verbose
Running external command line application(s). This may print messages to stdout and/or stderr.
The command(s) being run are below. These commands cannot be manually re-run as they will depend on temporary files that no longer exist.

Command: run_dada.R /var/folders/h3/2l54lt8d69db92g2mlsc_wnw0000gp/T/qiime2-archive-tg33bd_3/52802aaa-e515-4eb6-a193-4466da154458/data /var/folders/h3/2l54lt8d69db92g2mlsc_wnw0000gp/T/tmp83ajfcmt/output.tsv.biom 98 0 /var/folders/h3/2l54lt8d69db92g2mlsc_wnw0000gp/T/tmp83ajfcmt

Loading required package: Rcpp
There were 50 or more warnings (use warnings() to see the first 50)
Initial error matrix unspecified. Error rates will be initialized to the maximum possible estimate from this data.
Error in colSums(trans[paste0(nti, "2", c("A", "C", "G", "T")), ]) :
  'x' must be an array of at least two dimensions
Calls: dada -> errorEstimationFunction -> colSums
Execution halted
Traceback (most recent call last):
  File "/Users/jairideout/miniconda3/envs/qiime2-test/bin/qiime", line 6, in <module>
    sys.exit(q2cli.__main__.qiime())
  File "/Users/jairideout/miniconda3/envs/qiime2-test/lib/python3.5/site-packages/click/core.py", line 716, in __call__
    return self.main(*args, **kwargs)
  File "/Users/jairideout/miniconda3/envs/qiime2-test/lib/python3.5/site-packages/click/core.py", line 696, in main
    rv = self.invoke(ctx)
  File "/Users/jairideout/miniconda3/envs/qiime2-test/lib/python3.5/site-packages/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
[...]

Issue #12, opened 2016-10-03
Add ability to join paired reads


johnchase:
Similar to running dada2 in R it would be nice to have the ability to join paired reads with the q2 plugin

benjjneb:
Is there currently a QIIME object for paired fastq files (or lists of paired fastq files)?

johnchase:
I don't believe that there is. @ebolyen can you confirm?

ebolyen:
There is if I understand the question correctly: `SampleData[PairedEndSequencesWithQuality]` which uses `SingleLanePerSamplePairedEndFastqDirFmt` as its backing directory format.

benjjneb:
If that exists already, then adding a paired end workflow should be straightforward.

Is there documentation of `SampleData[PairedEndSequencesWithQuality]` objects? For example, can it be assumed there are always forward and reverse files for each sample?

ebolyen:
No real documentation, and we can't yet enforce that there *exists* a forward and reverse for each sample, but I believe it is our goal to be able to validate that property. That particular directory format does contain a manifest file which maps the sample id to the filepath and direction.

[relevant lines for format](https://github.com/qiime2/q2-types/blob/master/q2_types/per_sample_sequences/_format.py#L51-L76)
[manifest definition](https://github.com/qiime2/q2-types/blob/master/q2_types/per_sample_sequences/_format.py#L51-L76)
[q2-demux which manipulates these](https://github.com/qiime2/q2-demux)

These parts of the API (multi-file directory formats) are still very rough around the edges, but we're thinking about ways to make it easier.

benjjneb:
Added in a1857e54ef0599e956c2f73bed1b0bc6851ac946 and b66432da308a924a26f315a517c934b4a335af2f

Assumes dada2 1.2+

PR #34, opened 2017-01-16
Paired end workflow

benjjneb:
This branch implements, on the R side, an updated multithreaded workflow (`run_dada_faster.R`) that also estimates error rates from a subset of the data. The speedup relative to the old workflow will vary based on the size of the dataset and the computational resources available, but will increase for larger datasets and more available cores. On small datasets (eg. the mothur Miseq tutorial dataset) and a laptop the speedup is about 2x. On large datasets (eg. our Hiseq lanes with 800 multiplexed samples) and a 32-core compute node, the speedup is about 100x.

This branch also implements, on the R side, a similarly performant paired-end workflow (`run_dada_faster_paired.R`).

I've also tried to make the standard output somewhat more informative for this push-button application, and to improve the commented documentation of the R scripts.

The previous script (`run_dada.R`) remains unchanged (excepting comments) for testing purposes.

NOTHING WORKS FROM THE QIIME2/PYTHON SIDE YET! I've only updated the R scripts, and tested them with direct Rscript invocations. Corresponding changes are now needed on the python side.

This branch addresses #9, #12, #16 and #31. It depends on #33 as it expects dada2 version 1.2


jairideout:
Awesome, thanks @benjjneb! I'm assigning @ebolyen to this PR, he'll also be taking the lead on hooking up multithreading and paired-end support on the Python side of things.

gregcaporaso:
Closing as these commits are part of #36.

Issue #32, opened 2016-12-27
Expectations for primer removal in Q2?


benjjneb:
DADA2's chimera removal step requires primers to have been removed, as otherwise the ambiguous nucleotides in most primer sets cause large numbers of false-positive chimeras to be identified.

We expect the user to remove primers (if they are in their reads) prior to starting the R DADA2 workflows, and try to make that as clear as possible in our documentation. To what extent is primer removal to be expected in Q2 prior to the "denoise" step? Should a check-and-warn step for potential untrimmed primers be added to the q2-dada2 workflows?

gregcaporaso:
This plugin should assume that this has already happened. As it stands now, with the EMP protocol we already know that primers are not present. We'll add a warning to the documentation for now so that users are aware of this, and in the future we may add a semantic property to indicate that primers are remove (or more generally that the sequence is sequencing artifact free).

gregcaporaso:
I think we should keep this one open for now, until we figure out the exact plan for how to handle this and until we add a note to the docs about it.

ebolyen:
We are going to need something to trim reverse primers. Perhaps a plugin for (trimmomatic)[http://www.usadellab.org/cms/?page=trimmomatic] could be used. (Although we've never tried using Java in conda).

benjjneb:
You might want to look at [cutadapt](http://cutadapt.readthedocs.io/en/stable/index.html) too, as it is python and pip installable already.

ConstantinoSchillebeeckx:
Hi everyone, I've stumbled upon this thread after having run DADA2 with primers still present in my reads. 😓 

I'm looking at how to incorporate said trim step with `cutadapt` and am looking at doing the following:

1. `qiime demux emp-paired`
2. `qiime tools export` demux.qza
3. `cutadapt` on exported demuxed reads
4. `qiime tools import` trimmed reads 
5. `qiime dada2 denoise-paired`

Is there a more _qiime2-like_ way of doing this without having to do the import-trim-export steps? Ideally I'd like to run `cutadapt` directly on demuz.qza

---
Edit: looks like this is on [your radar already](https://forum.qiime2.org/t/its2-and-implementing-cutadapt/1700/3)? If so, consider this a +1

jairideout:
[...]

PR #36, opened 2017-01-27
Paired end python

ebolyen:
Based on #34 

I noticed that `maxEE` seems to be a float, so I changed the registration to indicate that. I'm not confident on the defaults (copied them from the R script) so if special attention could be paid to those, it would be greatly appreciated!


q2_dada2/_denoise.py
```diff
@@ -6,28 +6,93 @@
...
+            pass
+    if line_count < 2:
+        raise ValueError("No features remain after denoising. "
+                         "Try adjusting your arguments.")
+
+
+def denoise_single(demultiplexed_seqs: SingleLanePerSampleSingleEndFastqDirFmt,
+                   trunc_len: int, trim_left: int, max_ee: float=2.5,
+                   trunc_q: int=2, n_threads: int=1,
+                   n_reads_learn: int=1000000, hashed_feature_ids: bool=True
```
ebolyen: Not sure about the defaults.

q2_dada2/_denoise.py
```diff
@@ -6,28 +6,93 @@
...
+                (skbio.DNA(id_, metadata={'id': id_})
+                 for id_ in table.ids(axis='observation')))
+    return table, rep_sequences
+
+
+def denoise_paired(demultiplexed_seqs: SingleLanePerSamplePairedEndFastqDirFmt,
+                   trunc_len_f: int, trunc_len_r: int,
+                   trim_left_f: int, trim_left_r: int,
+                   max_ee: float=2.5, trunc_q: int=2, n_threads: int=1,
+                   n_reads_learn: int=1000000, hashed_feature_ids: bool=True
```
ebolyen: same here, what should the defaults be?

ebolyen:
Looks like I'm missing a recent commit, I'll rebase

q2_dada2/_denoise.py
benjjneb: I would probably default `max_ee` to `2` and `trim_left` to `0` (under the assumption discussed in #32 that primers are already trimmed).

Also, default for `n_threads` should be chosen based on what you all want as a default - single-threaded (`1`) or detect and use all available cores (`0`, and should explain meaning of that in user-facing docs).

q2_dada2/_denoise.py
benjjneb: As above.

q2_dada2/_denoise.py
ebolyen: Thanks @benjjneb! I think single-threaded is probably the way to go for the moment, and we should be able to tie this parameter into more interesting things in the future.

Is `n_reads_learn` a good value? Should it be `0`?

benjjneb: 1M is a good default for `n_reads_learn`.

Def not 0, that will cause the script to learn the error rates from just the first sample, which is way too few reads in the most common Miseq protocols.

ebolyen: So if I understand this correctly, 0 is all reads from only the first sample, and any other number is randomly sampled from all reads of all samples?

benjjneb: The script will use the first X samples, where X is at least 1, and the number of reads in X is at least greater than `n_reads_learn`, to learn the error rates.

So `n_reads_learn=0` will result in just 1 sample, the first sample, being used to estimate the error rates.

*`n_reads_learn` is really a minimum number of reads, not an exact number*

ebolyen: Interesting, thanks for the clarification!

q2_dada2/_denoise.py
```diff
@@ -6,28 +6,93 @@
...
-            trunc_len: int, trim_left: int, max_ee: int=2, truncq: int=2,
-            hashed_feature_ids: bool=True) -> (biom.Table, DNAIterator):
+def _check_featureless_table(fp):
+    with open(fp) as fh:
+        # There is a comment line and a header before the feature data
+        for line_count, _ in zip(range(3), fh):
+            pass
+    if line_count < 2:
+        raise ValueError("No features remain after denoising. "
+                         "Try adjusting your arguments.")
```
gregcaporaso: *arguments* to *truncation and trim parameter settings*

q2_dada2/_denoise.py
```diff
@@ -6,28 +6,93 @@
...
+                   trunc_len: int, trim_left: int=0, max_ee: float=2.0,
+                   trunc_q: int=2, n_threads: int=1,
+                   n_reads_learn: int=1000000, hashed_feature_ids: bool=True
+                   ) -> (biom.Table, DNAIterator):
     with tempfile.TemporaryDirectory() as temp_dir_name:
         biom_fp = os.path.join(temp_dir_name, 'output.tsv.biom')
-        cmd = ['run_dada.R', str(demultiplexed_seqs), biom_fp,
-               str(trunc_len), str(trim_left), str(max_ee), str(truncq),
-               temp_dir_name]
+        cmd = ['run_dada_faster.R',
```
gregcaporaso: Should we change the name of the R script to `run_data_single.R` (and `run_data_paired.R`)?

q2_dada2/_denoise.py
gregcaporaso: *arguments* to *truncation and trim parameter settings*

q2_dada2/_denoise.py
```diff
@@ -6,28 +6,93 @@
...
+                       for id_ in table.ids(axis='observation')}
+            table.update_ids(fid_map, axis='observation', inplace=True)
+
+            rep_sequences = DNAIterator((skbio.DNA(k, metadata={'id': v})
+                                         for k, v in fid_map.items()))
+        else:
+            rep_sequences = DNAIterator(
+                (skbio.DNA(id_, metadata={'id': id_})
+                 for id_ in table.ids(axis='observation')))
+    return table, rep_sequences
```
gregcaporaso: Can lines 44 - 64 be moved to a single function that can be called from `denoise_single` and `denoise_paired`?

q2_dada2/plugin_setup.py
```diff
@@ -32,25 +34,59 @@
...
+                'trunc_q': qiime2.plugin.Int,
+                'n_threads': qiime2.plugin.Int,
+                'n_reads_learn': qiime2.plugin.Int,
                 'hashed_feature_ids': qiime2.plugin.Bool},
     outputs=[('table', FeatureTable[Frequency]),
              ('representative_sequences', FeatureData[Sequence])],
-    name='Denoise and dereplicate',
-    description=('This method denoises sequences, dereplicates them, and '
-                 'filters chimeras.')
+    name='Denoise and dereplicate single-end',
```
gregcaporaso: *single-end sequences.*

PR #37, opened 2017-02-01
add input/output/parameter help text

gregcaporaso:
It probably makes sense to ~~not~~ squash these commits.

### DIFF
--- a/q2_dada2/plugin_setup.py
+++ b/q2_dada2/plugin_setup.py
@@ -71,14 +71,15 @@ plugin.methods.register_function(
         'n_reads_learn': ('The number of reads to use when training the '
                           'error model. Smaller numbers will result in a '
                           'shorter run time but a less reliable error '
-                          'model.'),
+                          'model. If 0 is provided, all input reads will be '
+                          'used.'),
         'hashed_feature_ids': ('If true, the feature ids in the resulting '
                                'table will be presented as hashes of the '
                                'sequences defining each feature. The hash '
                                'will always be the same for the same sequence '
                                'so this allows feature tables to be merged '
                                'across runs of this method. You should only '
-                               'merge tables if the exact same paraemters are '
+                               'merge tables if the exact same parameters are '
                                'used for each run.')
     },
     output_descriptions={
@@ -136,21 +137,24 @@ plugin.methods.register_function(
                    'value will be discarded.'),
         'trunc_q': ('Reads are truncated at the first instance of a quality '
                     'score less than or equal to this value. If the resulting '
-                    'read is then shorter than `trunc_len`, it is discarded.'),
+                    'read is then shorter than `trunc_len_f` or `trunc_len_r` '
+                    '(depending on the direction of the read) it is '
+                    'discarded.'),
         'n_threads': ('The number of threads to use for multithreaded '
                       'processing. If 0 is provided, all available cores will '
                       'be used.'),
         'n_reads_learn': ('The number of reads to use when training the '
                           'error model. Smaller numbers will result in a '
                           'shorter run time but a less reliable error '
-                          'model.'),
+                          'model. If 0 is provided, all input reads will be '
+                          'used.'),
         'hashed_feature_ids': ('If true, the feature ids in the resulting '
                                'table will be presented as hashes of the '
                                'sequences defining each feature. The hash '
                                'will always be the same for the same sequence '
                                'so this allows feature tables to be merged '
                                'across runs of this method. You should only '
-                               'merge tables if the exact same paraemters are '
+                               'merge tables if the exact same parameters are '
                                'used for each run.')
     },
     output_descriptions={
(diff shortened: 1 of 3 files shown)


q2_dada2/plugin_setup.py
```diff
@@ -46,6 +46,47 @@
...
+        'trunc_q': ('Reads are truncated at the first instance of a quality '
+                    'score less than or equal to this value. If the resulting '
+                    'read is then shorter than `trunc_len`, it is discarded.'),
+        'n_threads': ('The number of threads to use for multithreaded '
+                      'processing. If 0 is provided, all available cores will '
+                      'be used.'),
+        'n_reads_learn': ('The number of reads to use when training the '
+                          'error model. Smaller numbers will result in a '
+                          'shorter run time but a less reliable error '
+                          'model.'),
```
ebolyen: Do we want to mention the behavior of `0` or is it easier to omit it?

q2_dada2/plugin_setup.py
```diff
@@ -66,6 +107,60 @@
...
+                        'will be discarded.'),
+        'trim_left_r': ('Position at which reverse read sequences should be '
+                        'trimmed due to low quality. This trims the 5\' end '
+                        'of the input sequences, which will be the bases that '
+                        'were sequenced in the first cycles.'),
+        'max_ee': ('Reads with number of expected errors higher than this '
+                   'value will be discarded.'),
+        'trunc_q': ('Reads are truncated at the first instance of a quality '
+                    'score less than or equal to this value. If the resulting '
+                    'read is then shorter than `trunc_len`, it is discarded.'),
```
ebolyen: `trunc_len` -> `trunc_len_f or trunc_len_r (depending on sequence direction)`

q2_dada2/plugin_setup.py
```diff
@@ -66,6 +107,60 @@
...
+        'n_threads': ('The number of threads to use for multithreaded '
+                      'processing. If 0 is provided, all available cores will '
+                      'be used.'),
+        'n_reads_learn': ('The number of reads to use when training the '
+                          'error model. Smaller numbers will result in a '
+                          'shorter run time but a less reliable error '
+                          'model.'),
```
ebolyen: Same question about 0.

q2_dada2/plugin_setup.py
```diff
@@ -66,6 +107,60 @@
...
+        'hashed_feature_ids': ('If true, the feature ids in the resulting '
+                               'table will be presented as hashes of the '
+                               'sequences defining each feature. The hash '
+                               'will always be the same for the same sequence '
+                               'so this allows feature tables to be merged '
+                               'across runs of this method. You should only '
+                               'merge tables if the exact same paraemters are '
```
ebolyen: `parameters`

q2_dada2/plugin_setup.py
```diff
@@ -46,6 +46,47 @@
...
```
ebolyen: `parameters`

q2_dada2/plugin_setup.py
```diff
@@ -77,6 +172,13 @@
     inputs={'demultiplexed_seqs':
             SampleData[SequencesWithQuality | PairedEndSequencesWithQuality]},
     parameters={'n': qiime2.plugin.Int},
+    input_descriptions={
+        'demultiplexed_seqs': ('The demultiplexed sequences for which '
+                               'quality plots should be generated.')
+    },
+    parameter_descriptions={
+        'n': 'The number of per-sample quality plots to generate.'
```
ebolyen: Add `Twice the number of plots are generated in the case of paired-end sequences.`

q2_dada2/plugin_setup.py
gregcaporaso: This comment was on the wrong method - I updated it on the other method.

q2_dada2/plugin_setup.py
```diff
@@ -136,21 +137,24 @@
...
+                    'discarded.'),
         'n_threads': ('The number of threads to use for multithreaded '
                       'processing. If 0 is provided, all available cores will '
                       'be used.'),
         'n_reads_learn': ('The number of reads to use when training the '
                           'error model. Smaller numbers will result in a '
                           'shorter run time but a less reliable error '
-                          'model.'),
+                          'model. If 0 is provided, all input reads will be '
+                          'used.'),
```
ebolyen: `If 0 is provided, all input reads will be used from the first sample.`

https://github.com/qiime2/q2-dada2/pull/36#discussion_r98472656

q2_dada2/plugin_setup.py
```diff
@@ -71,14 +71,15 @@
...
```
ebolyen: `If 0 is provided, all input reads will be used from the first sample.`

https://github.com/qiime2/q2-dada2/pull/36#discussion_r98472656

ebolyen: Maybe we should update the R script as well to reflect the behavior better.

jairideout: 👍, can you also describe what the "first sample" is? I think it's alphabetical.

q2_dada2/plugin_setup.py
jairideout: same comment as https://github.com/qiime2/q2-dada2/pull/37#discussion_r99168285

q2_dada2/plugin_setup.py
gregcaporaso: I'm removing the description of the `0` behavior, since @benjjneb suggests that we wouldn't want to do this for a MiSeq run (which is most common for us). We're then going to require this value to be greater than zero in a separate PR where we include checks of these values.

PR #40, opened 2017-02-02
ENH: adds tests, arg validation, and filter errors

ebolyen:
- [x] Validates input domain.
- [x] Uses return code to detect completely filtered reads.
- [x] Adds tests for single end
- [x] Adds tests for paired end

~~** **DO NOT MERGE** **~~


gregcaporaso:
@ebolyen, is travis using the right version of DADA2 here, or do we need to follow @jairideout's new install instructions?

jairideout:
Looks great, thanks @ebolyen! Pair-reviewed with @thermokarst and @ebolyen.

Issue #50, opened 2017-03-12
Add chimera filtering options


benjjneb:
In my opinion the options related to chimera removal are important to expose to users of the plugin. This is especially true because dada2 implements two kinds of chimera removal (pooled and consensus), and as of our next 1.4 release we will be officially recommending consensus removal for large datasets (default is pooled).

gregcaporaso:
Is this something that could happen with the current release, or does it depend on the 1.4 release of DADA2?

benjjneb:
Can be done now, but multithreading will only work on pooled chimera removal until 1.4.

I'll make a branch with an updated R script.

benjjneb:
Added in 2017.4

PR #53, opened 2017-03-31
Chimera options

benjjneb:
This pull request addresses #50 Two options were added.

`chimera-method`: valid options are `none`, `pooled` and `consensus` (where `pooled` is the current behavior).

`min-parent-abundance`: A numeric value which should be 1.0 or greater.

I've added the options to the R scripts and verified thy work. I also added in python code that passes the CI tests, but needs to be reviewed and probably fixed. The python side has not been tested. Possibly a test should be added as well?

### DIFF
--- a/q2_dada2/_denoise.py
+++ b/q2_dada2/_denoise.py
@@ -33,6 +33,7 @@ def _check_featureless_table(fp):
 
 _WHOLE_NUM = (lambda x: x >= 0, 'non-negative')
 _NAT_NUM = (lambda x: x > 0, 'greater than zero')
+_CHIM_STR = (lambda x: x in {'pooled', 'consensus', 'none'}, 'pooled, consensus or none')
 # Better to choose to skip, than to implicitly ignore things that KeyError
 _SKIP = (lambda x: True, '')
 _valid_inputs = {
@@ -44,7 +45,7 @@ _valid_inputs = {
     'trim_left_r': _WHOLE_NUM,
     'max_ee': _NAT_NUM,
     'trunc_q': _WHOLE_NUM,
-    'chimera_method': (lambda x: x in {'pooled', 'consensus', 'none'}),
+    'chimera_method': _CHIM_STR,
     'min_parent_abundance': _NAT_NUM,
     'n_threads': _WHOLE_NUM,
     # 0 is technically allowed, but we don't want to support it because it only
--- a/q2_dada2/_denoise.py
+++ b/q2_dada2/_denoise.py
@@ -44,6 +44,8 @@ _valid_inputs = {
     'trim_left_r': _WHOLE_NUM,
     'max_ee': _NAT_NUM,
     'trunc_q': _WHOLE_NUM,
+    'chimera_method': _SKIP, # TEST FOR VALIDITY HERE? Value should be 'pooled', 'consensus' or 'none'
+    'min_parent_abundance': _NAT_NUM,
     'n_threads': _WHOLE_NUM,
     # 0 is technically allowed, but we don't want to support it because it only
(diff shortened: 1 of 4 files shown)


q2_dada2/_denoise.py
```diff
@@ -44,6 +44,10 @@ def _check_featureless_table(fp):
     'trim_left_r': _WHOLE_NUM,
     'max_ee': _NAT_NUM,
     'trunc_q': _WHOLE_NUM,
+    'chimera_method': _SKIP,
+    # TEST chimera_method FOR VALIDITY HERE?
+    # Value should be 'pooled', 'consensus' or 'none'
```
ebolyen: We can skip this if we add a predicate to the semantic type (I'll comment on that further down).

q2_dada2/plugin_setup.py
```diff
@@ -41,6 +41,9 @@
                 'trim_left': qiime2.plugin.Int,
                 'max_ee': qiime2.plugin.Float,
                 'trunc_q': qiime2.plugin.Int,
+                'chimera_method': qiime2.plugin.Str,
+                # IS qiime2.plugin.Str THE RIGHT TYPE?
```
ebolyen: Yup!

This can be:
```python
qiime2.plugin.Str % qiime2.plugin.Choices({'pooled', 'consensus', 'none'})
```

We also now have better support for `qiime2.plugin.Range` for the other number primitives in the future.

q2_dada2/plugin_setup.py
```diff
@@ -65,6 +68,20 @@
         'trunc_q': ('Reads are truncated at the first instance of a quality '
                     'score less than or equal to this value. If the resulting '
                     'read is then shorter than `trunc_len`, it is discarded.'),
+        'chimera_method': ('The method used to remove chimeras. '
+                           'none: No chimera removal is performed. '
+                           'pooled: All reads are pooled prior to chimera '
+                           'detection. consensus: Chimeras are detected in '
+                           'samples individually, and sequences found '
+                           'chimeric in a sufficient fraction of samples are '
+                           'removed.'),
```
ebolyen: I would quote the values like `"none"`, `"pooled"`, `"consensus"`.

q2_dada2/plugin_setup.py
```diff
@@ -102,6 +119,9 @@
                 'trim_left_r': qiime2.plugin.Int,
                 'max_ee': qiime2.plugin.Float,
                 'trunc_q': qiime2.plugin.Int,
+                'chimera_method': qiime2.plugin.Str,
```
ebolyen: same as above

q2_dada2/_denoise.py
ebolyen: If you wanted to validate independent of the semantic type, you could add: 
`'chimera_method': lambda x: x in {'pooled', 'consensus', 'none'},`.

But I think ideally this `_valid_inputs` dict would be removed entirely in the future (using `qiime2.plugin.Range`).

gregcaporaso:
@benjjneb, think you'll be able to address @ebolyen's comments by Friday? We're hoping to have all new code in by Friday, so we can release next week. Let us know if you need any help with meeting that timeline.

benjjneb:
A miracle occurred and travis accepted my white space!

I added in @ebolyen suggestions, but someone should probably take a quick look before merging this in.

ebolyen:
I'm going to merge this and write some simple unit-tests exercising that the parameters are passed through. @benjjneb, if you have any specific edge-cases worth testing, I'd be happy to include them, otherwise it's just going to use our oversimple mock data.

PR #54, opened 2017-04-11
Add smoketests and rename parameter

ebolyen:
Don't squash these commits.

This was a little less useful than I was hoping, as only the single-end data was really impacted by changing the chimera_method, and min_fold_over_parent_abundance had no impact in any situation. I think this is a case where we would need hand-curated test-data instead of heavily subsampled tutorial data.


ebolyen:
Thanks @benjjneb! I completely typo-ed that name and then kept reusing the name in my buffer. I'll fix that.

Also, I've set the value as high as `100` and saw no effect. I'm guessing there aren't really any chimeric sequences in our over-subsampled data.

ebolyen:
Fixed the parameter names!

q2_dada2/plugin_setup.py
```diff
@@ -77,13 +77,13 @@
...
-                                 'Values should be greater than or equal to 1 '
-                                 '(i.e. parents should be more '
-                                 'abundant than the sequence being tested).'),
+        'min_fold_parent_over_abundance': (
+            'The minimum abundance of potential parents of a sequence being '
+            'tested as chimeric, expressed as a fold-change versus the '
+            'abundance of the sequence being tested. Values should be greater '
+            'than or equal to 1 (i.e. parents should be more abundant than '
+            'the sequence being tested). This parameter has no effect if '
+            'chimera_method is "none".'),
```
ebolyen: This diff is kind of mangled, but I just wanted to draw attention to the last line here (which is the only actual change).

benjjneb: Looks good to me.

gregcaporaso: This makes sense to me. One question though - should it be a requirement that the value is greater than or equal to 1?

benjjneb: >should it be a requirement that the value is greater than or equal to 1?

Probably, although we don't check for >=1 in the R package either right now. In practice very few people set this parameter, but I thought it was important to include in light of the discussion in Edgar's unoise2 paper about whether the min_fold should be 1 or 2.

gregcaporaso: It's easy to do that check, so I vote that we do it (maybe it'll result in one less support question...).

ebolyen: Just FYI, the plugin is already range-checking it's inputs. So what's documented is what's asserted right now.

Issue #56, opened 2017-05-01
Use DADA2 1.4 to fix quality-score precedence bug in parameters


ebolyen:
It looks like this is up on bioconda, although it lists both `1.4` and `1.4.0` and I haven't had a chance to figure out the difference. 

From a [forum post](https://forum.qiime2.org/t/problem-with-dada2-denoise-paired-command/520/12?u=ebolyen):

> There is a small bug in the 1.2 version of the DADA2 package the plugin is using (minQ is enforced before trimming, rather than after) that is preventing p-trim-left-f from working as expected. So you would need to trim off that bad staring base position with another bit of software to get it to work with the current QIIME2 plugin. When the plugin upgrades to the 1.4 version of DADA2 the p-trim-left-f approach should work.

The conda-recipe pins the version to 1.2 so this will need to wait until 2017.5 unless we want to patch.

@benjjneb would it be more appropriate to list our dependency as `>= 1.2` (or rather `>= 1.4` now) so that just upgrading `dada2` fixes these kinds of issues?

benjjneb:
My concern about removing the fixed version dependence is that I haven't guaranteed that the current q2-dada2 R scripts are robust to all default parameter changes, and there were 2 defaults changed in our 1.4 release.

I think this bug is sufficiently minor to leave until another release of the plugin, at which point it might be worth reviewing the R code and explicitly defining every parameter to allow for `>= 1.4` dependencies in the future.

ebolyen:
fixed by #61

PR #61, opened 2017-05-18
MAINT: update to dada2 1.4.0

### LINKED ISSUE #57: Update plugin to 1.4.0 R package
The new version is available via bioconda (version 1.4.0). 

This is a precursor to fixing #56 

The new package version allows variable length amplicons, and once available #43 and #52 should be addressed.

ebolyen:
Fixes #57 

Tests won't pass yet (a default changed, without it they pass just fine). 

I just want to check that I caught the important updates from 1.2 to 1.4.

I'm getting a bunch of warnings from other packages, but they've been around a while I think. Not sure how I would resolve them.
```
Warning messages:
1: multiple methods tables found for ‘arbind’
2: multiple methods tables found for ‘acbind’
3: replacing previous import ‘IRanges::arbind’ by ‘SummarizedExperiment::arbind’ when loading ‘GenomicAlignments’
4: replacing previous import ‘IRanges::acbind’ by ‘SummarizedExperiment::acbind’ when loading ‘GenomicAlignments’
4: replacing previous import ‘IRanges::acbind’ by ‘SummarizedExperiment::acbind’ when loading ‘GenomicAlignments’
5: multiple methods tables found for ‘left’
6: multiple methods tables found for ‘right’
```


benjjneb:
Two questions: 

Why `r3.3.1_0` in the yaml? The current R version is 3.4.0 and that is the version officially supported by the Bioconductor release.

Related, how can I test drive the precise installation this will produce on my own machine? Are there conda commands I can call by hand to make the same result?

At first glance this looks OK, but I want to double-check a couple things. How tight is the schedule to include in the next release? If its really soon, it may be better to hold off merging this until next month's release. Plus I'll have time to have added in the variable-amplicon/ITS support by then as well.

ebolyen:
>Why r3.3.1_0 in the yaml? 

That's the build number of the package. I ran into an issue where conda would consider `1.4` and `1.4.0` to be semantically identical (they are under PEP440) and so would randomly switch which one it used. The build number fixes it to a single file. As for why it says `r3.3.1_0` it means this build was compiled against `r 3.3.1` and is build `0`. I can't tell you why it is compiled against that particular R, but these are the only versions available:
```
10:35:19 ~ § conda search -c bioconda bioconductor-dada2
Fetching package metadata ...........
bioconductor-dada2           0.10.0                        0  bioconda        
                             0.10.1                        0  bioconda        
                             0.10.3                        0  bioconda        
                             0.10.4                        0  bioconda        
                             0.99.10                       0  bioconda        
                             1.0.3                         0  bioconda        
                             1.2                    r3.3.1_0  bioconda        
                             1.4                    r3.3.1_0  bioconda        
                             1.4.0                  r3.3.1_0  bioconda   
```

> Are there conda commands I can call by hand to make the same result?

Yup! (I'm going to assume you're using `osx` if not swap the url to `linux`). This is basically what I did:
```
conda create -n test-dada2-1.4.0 --file https://raw.githubusercontent.com/qiime2/environment-files/master/latest/staging/qiime2-latest-conda-osx-64.txt
source activate test-dada2-1.4.0
pip install <your clone of this repo>
conda install -c bioconda bioconductor-dada2=1.4.0=r3.3.1_0
```

> How tight is the schedule to include in the next release?

We're hoping to get things *merged* by the 22nd/23rd. This would fix #56 so that's why we're interested in getting it into this release if we can.

Thanks @benjjneb!

benjjneb:
Thanks @ebolyen 

Testing looks good from my end. The defaults that changed in the 1.4 R package were already explicitly specified in the Q2 R scripts so no problems there. So I would go ahead and merge if the tests are passing.

One big mystery this revealed... the dada2 scripts appear to run about **5x slower (!!)** w/in the Q2 conda environment than when just using my native R installation. I'll open a separate issue on that though.

Note: I had to add `conda uninstall q2-dada2` before installing the dev dada2 version and bioconductor package, or I got an UnsatisfiableError conflict between 1.2 and 1.4. (And for my own notes, also needed `conda install -c conda-forge pandoc=1.19.2` to get rmarkdown working)

ps: In your first post you mentioned some warning messages (eg. `1: multiple methods tables found for ‘arbind’`). Those are safe to ignore. That's just R complaining about multiple packages defining functions with the same name, which should only be a problem in interactive usage.

ebolyen:
Thanks @benjjneb, I've updated the tests to expect the new default. 

(I'm amazed out how stable dada2's results are generally, switching from pooled to consensus just added a new SV but the counts of everything else were the same!)

ebolyen:
Oh, you know what. The tests are going to fail until our environment files update to the new dependency. The tests pass locally with `dada2 1.4.0`

Issue #52, opened 2017-03-30
Add truncLen requirements to the denoise-paired help text


benjjneb:
For the F/R reads to be successfully merged, `trunc-len-f` + `trunc-len-r` must be greater than the length of the amplicon + 20 nucleotides (the 20 nts is the length of the overlap).

This requirement should be reflected in the documentation for `denoise-paired`.

maxvonhippel:
Note that in the code these appear as `truncLenF` and `truncLenR`.

Also, I'm working on this now.

jairideout:
Fixed in #63

PR #63, opened 2017-05-23
Improved truncLen error message

maxvonhippel:
This PR addresses issue #52 

The new (hopefully improved) error message from `denoise_paired` in the event that `run_dada_paired.R` fails says the following:

````PYTHON
raise ValueError(
                    "No reads passed the filter. trunc_len_f (%r) or"
                    " trunc_len_r (%r) may be individually longer than"
                    " read lengths, or trunc_len_f + trunc_len_r may be"
                    " shorter than the length of the amplicon + 20"
                    " nucleotides (the length of the overlap). Alternatively,"
                    " other arguments (such as max_ee or trunc_q) may be"
                    " preventing reads from passing the filter."
                    % (trunc_len_f, trunc_len_r))
````

### DIFF
--- a/q2_dada2/_denoise.py
+++ b/q2_dada2/_denoise.py
@@ -177,7 +177,10 @@ def denoise_paired(demultiplexed_seqs: SingleLanePerSamplePairedEndFastqDirFmt,
             if e.returncode == 2:
                 raise ValueError(
                     "No reads passed the filter. trunc_len_f (%r) or"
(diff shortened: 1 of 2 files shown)


ebolyen:
I like the error update, but we should also add this note to the documentation for `trunc_len_f/r` (in `plugin_steup.py`).

maxvonhippel:
@ebolyen right, good point!  I'll do that, thanks!

Issue #64, opened 2017-05-26
Add support for variable amplicon lengths


benjjneb:
Now that the plugin has been updated to 1.4, variable length amplicons can be supported, but the R script needs to be updated. Should be doable for the next release.

My first thought is to allow `trunc-len-f = 0` to be a special value that turns off the truncation length.

ebolyen:
Fixed in #66

PR #66, opened 2017-06-05
Variable length amplicons

benjjneb:
This adds support for variable length amplicons #64 

It also adds the multithreaded filtering implemented in dada2 package version 1.4. However, that part is causing some Travis failures I'm having trouble figuring out...

### DIFF
--- a/q2_dada2/assets/run_dada_paired.R
+++ b/q2_dada2/assets/run_dada_paired.R
@@ -44,10 +44,12 @@
 #
 # 6) truncLenF - The position at which to truncate forward reads. Forward reads shorter
 #               than truncLenF will be discarded.
+#               Special values: 0 - no truncation or length filtering.
 #    Ex: 240
 #
 # 7) truncLenR - The position at which to truncate reverse reads. Reverse reads shorter
 #               than truncLenR will be discarded.
+#               Special values: 0 - no truncation or length filtering.
 #    Ex: 160
 #
 # 8) trimLeftF - The number of nucleotides to remove from the start of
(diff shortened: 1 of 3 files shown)


benjjneb:
This adds support for variable length amplicons. Setting trunc-len/trunc-len-f/trunc-len-r to zero turns off truncation, and everything else proceeds as normal.

Truncation to a common length is still recommended for most use-cases (eg. 16S) but for loci like ITS it is sometimes strongly desirable to allow amplicons of variable length, and that should now be supported.

q2_dada2/assets/run_dada_single.R
```diff
@@ -188,10 +188,8 @@ seqtab <- makeSequenceTable(dds)
 
 # Remove chimeras
 cat("4) Remove chimeras (method = ", chimeraMethod, ")\n", sep="")
-if(chimeraMethod == "pooled") {
-  seqtab <- removeBimeraDenovo(seqtab, method=chimeraMethod, minFoldParentOverAbundance = minParentFold, multithread=multithread)
-} else if(chimeraMethod == "consensus") {
-  seqtab <- removeBimeraDenovo(seqtab, method=chimeraMethod, minFoldParentOverAbundance = minParentFold, multithread=multithread)
+if(chimeraMethod %in% c("pooled", "consensus")) {
```
ebolyen: > `%in%`

That's a very interesting spelling.

Thanks! This is much cleaner than what I had done to your R code.

q2_dada2/assets/run_dada_paired.R
```diff
@@ -44,10 +44,12 @@
 #
 # 6) truncLenF - The position at which to truncate forward reads. Forward reads shorter
 #               than truncLenF will be discarded.
+#               Special values: 0 - no truncation or length filtering.
 #    Ex: 240
 #
 # 7) truncLenR - The position at which to truncate reverse reads. Reverse reads shorter
 #               than truncLenR will be discarded.
+#               Special values: 0 - no truncation or length filtering.
```
ebolyen: Looks like `0` is already a supported value by the code! :+1:

PR #75, opened 2017-09-19
MAINT: update version pin

ebolyen:


ebolyen:
Needed for the transition to `conda-forge`'s R.

PR #82, opened 2018-01-04
BUG: change SSE mode to avoid segfault

### LINKED ISSUE #80: Segfault errors in 2017.12 plugin version
Adding this as a tracking issue for a seemingly related group of errors that popped up after the plugin was updated to use the current version of the R package (1.6) in Q2 2017.12

https://forum.qiime2.org/t/dada2-error-caught-segfault/2322
https://forum.qiime2.org/t/dada2-exit-code-11/2333
https://forum.qiime2.org/t/dada2-errors-return-code-1-and-11/2380
https://forum.qiime2.org/t/dada2-11-exit-code-paired-end-reads/2360

In every case, these users are seeing the following:

```
Plugin error from dada2:
  An error was encountered while running DADA2 in R (return code -11), please inspect stdout and stderr to learn more.
```

And in the log file...

```
2) Learning Error Rates
Initializing error rates to maximum possible estimate.
Sample 1 - 19785 reads in 10633 unique sequences.
*** caught segfault ***
address 0x8, cause ‘memory not mapped’
```

This indicates the segfault error is happening the first time that the `dada` function in the R package is called, and that some other functions (such as filtering) are working OK.

Also in all cases reported thus far, the affected users are running OSX, although several different versions have been reported so far:

```
I am using macOS High Sierra Version 10.13.1
I am using OX X Yosemite Version 10.10.5
I have macOS High Sierra Version 10.13.2.
I am running High Sierra 10.13.1.
```

ebolyen:
fixes #80 

@benjjneb, we really like the idea of having SSE still, so we're fine with moving forward with this. We'll wait on your go-ahead though since you are still testing stuff.

Thanks!

### DIFF
--- a/q2_dada2/assets/run_dada_paired.R
+++ b/q2_dada2/assets/run_dada_paired.R
@@ -196,9 +196,9 @@ for(i in seq_along(filtsF)) {
 drpsF <- drpsF[1:i]
 drpsR <- drpsR[1:i]
 cat("2a) Forward Reads\n")
-ddsF <- dada(drpsF, err=NULL, selfConsist=TRUE, multithread=multithread, VECTORIZED_ALIGNMENT=FALSE, SSE=2)
+ddsF <- dada(drpsF, err=NULL, selfConsist=TRUE, multithread=multithread, VECTORIZED_ALIGNMENT=FALSE, SSE=1)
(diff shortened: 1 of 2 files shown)


ebolyen:
Also this is good for merge. So feel free to hit the button whenever you are ready.

PR #83, opened 2018-01-11
MAINT: update copyright year; travis

ebolyen:


ebolyen:
So it looks like this has something to do with `mcmapply` and travis, don't know much more than that yet.

PR #91, opened 2018-03-09
ENH: denoise_pyro

### LINKED ISSUE #76: adding 454/ion torrent support
I think this is just a matter of exposing the parameters described [here](https://benjjneb.github.io/dada2/faq.html#can-i-use-dada2-with-my-454-or-ion-torrent-data). I can take this one (either for 2017.10 or 2017.11), and will do some research when I work on it to figure out if it makes sense as part of an existing method or a new one or two.

thermokarst:
Fixes #76


thermokarst:
@benjjneb, would you be interested or available for reviewing this PR?

q2_dada2/plugin_setup.py
```diff
@@ -212,3 +212,85 @@
...
+                              'merge tables if the exact same parameters are '
+                              'used for each run.'
+    },
+    output_descriptions={
+        'table': 'The resulting feature table.',
+        'representative_sequences': 'The resulting feature sequences. Each '
+                                    'feature in the feature table will be '
+                                    'represented by exactly one sequence.'
+    },
+    name='Denoise and dereplicate single-end sequences',
```
benjjneb: Maybe sequences -> pyrosequences here?

q2_dada2/_denoise.py
```diff
@@ -105,23 +108,28 @@ def _denoise_helper(biom_fp, hashed_feature_ids):
     return table, rep_sequences
 
 
-def denoise_single(demultiplexed_seqs: SingleLanePerSampleSingleEndFastqDirFmt,
-                   trunc_len: int, trim_left: int=0, max_ee: float=2.0,
-                   trunc_q: int=2, chimera_method: str='consensus',
-                   min_fold_parent_over_abundance: float=1.0, n_threads: int=1,
-                   n_reads_learn: int=1000000, hashed_feature_ids: bool=True
-                   ) -> (biom.Table, DNAIterator):
+def _denoise_single(demultiplexed_seqs, trunc_len, trim_left, max_ee, trunc_q,
```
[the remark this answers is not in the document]
thermokarst: Ah, this is actually now a helper function - the diff is a little gross here - but for the sake of keeping things DRY this was split out into a reusable util. Default values are still set in the parent methods that will call into this function. I will update this with a comment to explain that more clearly. Thanks!

q2_dada2/_denoise.py
```diff
@@ -105,23 +108,28 @@ def _denoise_helper(biom_fp, hashed_feature_ids):
...
+
     with tempfile.TemporaryDirectory() as temp_dir_name:
         biom_fp = os.path.join(temp_dir_name, 'output.tsv.biom')
         cmd = ['run_dada_single.R',
                str(demultiplexed_seqs), biom_fp, temp_dir_name,
                str(trunc_len), str(trim_left), str(max_ee), str(trunc_q),
-               str(chimera_method), str(min_fold_parent_over_abundance),
-               str(n_threads), str(n_reads_learn)]
+               'Inf', str(chimera_method), str(min_fold_parent_over_abundance),
+               str(n_threads), str(n_reads_learn), homopolymer_gap_penalty,
```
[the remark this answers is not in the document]
thermokarst: Related to https://github.com/qiime2/q2-dada2/pull/91#discussion_r174300416. This function is only exposed to the method developers, so it assumes that they are passing in already valid params. But, given that people still make mistakes, I will update this to cast to `str`, just to make sure!

q2_dada2/_denoise.py
```diff
@@ -138,6 +151,18 @@ def denoise_single(demultiplexed_seqs: SingleLanePerSampleSingleEndFastqDirFmt,
...
 
 
+def denoise_single(demultiplexed_seqs: SingleLanePerSampleSingleEndFastqDirFmt,
+                   trunc_len: int, trim_left: int=0, max_ee: float=2.0,
+                   trunc_q: int=2, chimera_method: str='consensus',
+                   min_fold_parent_over_abundance: float=1.0, n_threads: int=1,
+                   n_reads_learn: int=1000000, hashed_feature_ids: bool=True
+                   ) -> (biom.Table, DNAIterator):
+    return _denoise_single(demultiplexed_seqs, trunc_len, trim_left, max_ee,
+                           trunc_q, 0, chimera_method,
```
gregcaporaso: This is a little nit-picky, but would you mind specifying these parameters as keyword arguments (e.g., `trunc_q=trunc_q, max_len=0, ...`). I like to do this when there are a lot of parameters for readability (e.g., helps to understand why one of these is always set to `0`), and to avoid a mix-up in the future if new parameters are added to the function that is being called. This would go for the two calls to `_denoise_single`.

q2_dada2/plugin_setup.py
```diff
@@ -212,3 +212,85 @@
...
+        'trim_left': 'Position at which sequences should be trimmed due to '
+                     'low quality. This trims the 5\' end of the '
+                     'of the input sequences, which will be the bases that '
+                     'were sequenced in the first cycles.',
+        'max_ee': 'Reads with number of expected errors higher than this '
+                  'value will be discarded.',
+        'trunc_q': 'Reads are truncated at the first instance of a quality '
+                   'score less than or equal to this value. If the resulting '
+                   'read is then shorter than `trunc_len`, it is discarded.',
+        'max_len': 'Remove raw reads which are longer than this value. If 0 '
```
gregcaporaso: It's not clear what _raw reads_ means here - could this just be _reads_ or _reads prior to trimming or truncation_

thermokarst:
Thanks for testing and reviewing, @gregcaporaso! I have pushed up some changes addressing your suggestions. Thanks!

gregcaporaso:
Thanks @thermokarst and @benjjneb!

PR #103, opened 2018-10-11
Update DADA2 1.8.0

ebolyen:
It looks like an OS X build was pushed for 2.48.0 (but not for linux).
Since bioconda doesn't have a way to synchronize the bionconductor
packages the bionconductor-dada2 package is pulling in the latest
biostrings which won't work as that package hasn't been rebuilt yet with
the namespace changes found here:

https://github.com/benjjneb/dada2/commit/85846e89bce8970a920bb492ae5108e79b8ab395#diff-7347fe5a0f184f79ef064e92e3beb297

This should fix https://busywork.qiime2.org/teams/main/pipelines/master-distribution/jobs/integration/builds/16 in the interim.

### DIFF
--- a/ci/recipe/meta.yaml
+++ b/ci/recipe/meta.yaml
@@ -21,11 +21,7 @@ requirements:
     - python 3.5*
     - setuptools
     - biom-format >=2.1.5,<2.2.0
-    # This pin only exists because namespaces where shuffled around and the
-    # next bioconductor version was uploaded only on OS X and only for
-    # biostrings at the moment.
-    - bioconductor-biostrings 2.48.0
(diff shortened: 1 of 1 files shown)


ebolyen:
Actually, we might hold off on this: https://github.com/bioconda/bioconda-recipes/issues/11315

It sounds like we might have the latest DADA2 available shortly which would be way better :)

ebolyen:
New DADA2 is live! Ready for merge.

benjjneb:
Hooray! When I get time (hopefully in Nov) I should be able to push script updates that close about half the open issues now that 1.8 is in.

PR #105, opened 2018-11-01
TST: Updating test data for dada2 1.8

thermokarst:


thermokarst:
Travis should fail, since it will be running an older release of DADA2

thermokarst:
@benjjneb - the results for these toy datasets appear to be fairly different when compared to previous versions of DADA2 - is this to be expected? Glancing at the [changelog](https://benjjneb.github.io/dada2/ReleaseNotes_1_8.html), it looks like there are some changes to defaults, etc which *seem* to explain this, but looking for a reality check from you. Thanks!

benjjneb:
I think it would be best to let me go through and review the scripts and output to ensure they work appropriately with 1.8, in part because of a couple of changed defaults. It will also give me a chance to update the scripts to close the various issues on the Issue tracker that were waiting on 1.8.

Unfortunately, I don't think I will get this done in time for the next Q2 release (end of next week as per Greg?), and if I'm being realistic I most likely will not get to this until Dec after my teaching is done.

Apologies for holding this up, and thanks so much for doing the work updating to the 1.8 build and navigating the other versioning/pinning issues, but I think it would be best to wait on pushing this branch to "production" until I have a chance to really give it a once over.

thermokarst:
Thanks @benjjneb - no worries, this all sounds great to me. Yes, the next release of QIIME 2 (2018.11) is slated for next week. After that, 2019.1 is on the books for the end of January 2019. Let us know what we can help with - you know where to find us! Thanks!

PR #112, opened 2019-02-18
Update to dada2 1.8.0; drop osx pins

epruesse:
BTW: `1.2.0a0` is `<` than `1.2.0`, so pinning should use `>=1.1.23,<1.2a0` or similar. Just in case someone pushes alphas.

If in doubt, use 

```
from conda.exports import VersionOrder
[str(x) for x in sorted(VersionOrder(v) for v in 
('1.2.0', '1.2.0a0', '1.2a0', '1.2', '1.1.99', '1.2.1', 
'1.2p1', '1.2post0', '1.2.12', '1.2.0.post0', '1.2.p0'))]
```

```
['1.1.99', '1.2a0', '1.2p1', '1.2.0a0', '1.2.p0', '1.2.0', 
'1.2', '1.2.0.post0', '1.2.1', '1.2.12', '1.2post0']
```

to verify what will actually happen. It's mostly PEP440 but not completely so, and not really intuitive (but then the intuition on that topic varies). 

I really like `1.2.0 < 1.2 < 1.2.1` ...

### DIFF
--- a/ci/recipe/meta.yaml
+++ b/ci/recipe/meta.yaml
@@ -19,27 +19,8 @@ requirements:
 
   run:
     - python {{ python }}
-    - biom-format >=2.1.5,<2.2.0
-    # These pins only exist because namespaces where shuffled around and the
-    # next bioconductor version was uploaded only on OS X and only for
-    # biostrings at the moment.
-    - bioconductor-biobase 2.38.0
-    - bioconductor-biocgenerics 0.24.0
-    - bioconductor-biocparallel 1.12.0
-    - bioconductor-biostrings 2.46.0
-    - bioconductor-dada2 1.6.0 r3.4.1_0
-    - bioconductor-delayedarray 0.4.1
-    - bioconductor-genomeinfodb 1.14.0
-    - bioconductor-genomeinfodbdata 1.0.0
-    - bioconductor-genomicalignments 1.14.1
-    - bioconductor-genomicranges 1.30.3
-    - bioconductor-iranges 2.12.0
-    - bioconductor-rsamtools 1.30.0
-    - bioconductor-s4vectors 0.16.0
-    - bioconductor-shortread 1.36.0
-    - bioconductor-summarizedexperiment 1.8.0
-    - bioconductor-xvector 0.18.0
-    - bioconductor-zlibbioc 1.24.0
+    - biom-format >=2.1.5,<2.2.0a0
+    - bioconductor-dada2 >=1.8.0,<1.9a0
     # openjdk is not a real dependency, but, r-base has a post-link and post-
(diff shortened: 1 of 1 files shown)


epruesse:
ping @thermokarst 

See https://github.com/qiime2/q2-alignment/pull/62

thermokarst:
@benjjneb - any reason to hold off on updating the plugin to DADA2 1.8? I recall we discussed in at least one other issue thread that there were some planned script updates to go with this, but I am not sure if those are necessary, or just nice-to-have.

epruesse:
I can change this back to 1.6 - just want to know whether busywork will come up with a plan that has the right `glib` so that SINA will comply on OSX.

thermokarst:
Let's stick with 1.6 for now.

benjjneb:
> @benjjneb - any reason to hold off on updating the plugin to DADA2 1.8?

Adding the script updates in is a reason, but perhaps more importantly is that I know there is one edge-case bug that will arise in the current q2-dada2 scripts once it updates to 1.8, which needs a workaround added to the plugin R scripts.

BTW, am I right to think that the next release of Q2 will be around the end of April?

thermokarst:
> Adding the script updates in is a reason, but perhaps more importantly is that I know there is one edge-case bug that will arise in the current q2-dada2 scripts once it updates to 1.8, which needs a workaround added to the plugin R scripts.

👍 @epruesse --- let's keep it at 1.6 for now, let's just drop the bioconductor pins.

> BTW, am I right to think that the next release of Q2 will be around the end of April?

Yep! Project page has specific dates for the next two releases: https://github.com/orgs/qiime2/projects

epruesse:
@thermokarst done

epruesse:
Uh-oh ... lots of failures: https://busywork.qiime2.org/teams/main/pipelines/master-distribution/jobs/unit-test-q2-dada2/builds/33

epruesse:
This looks like it doesn't like the biostrings version installed.

BTW: We can re-build older versions of dada2 on Bioconda, too. Just need to add `bioconductor-dada2/1.6.0/...`. 

BTW: If we need a rebuilt version of `bioconductor-dada2` - perhaps with im

thermokarst:
@epruesse --- see this PR for context: https://github.com/qiime2/q2-dada2/pull/103

We won't be able to drop these pins until we are able to update DADA2 to 1.8. I will revert these changes. Until then, you can run an env without the DADA2 stack in it, check out this gist from @ebolyen: https://gist.github.com/ebolyen/7cff6868d1fac82f6b5dae7a70f15858

@benjjneb --- anything we can do on this end to help? I am happy to take a stab at a PR, just let me know.

PR #120, opened 2019-08-19
IMP: Add new percentage-merged column to denoise-paired output

### LINKED ISSUE #115: denoise-paired: add % merged to stats output
**Improvement Description**
For paired-end, successful merging is crucial, and it would be useful to display the % merged. IMO the other columns are fine with just absolute values, bit relative yields are very important after merging: it is useful to know whether, e.g., 90% of sequences are failing to merge in specific samples and not others.

**Current Behavior**
Only absolute values are given.

**Proposed Behavior**
Provide additional column: merged % = (merged column / denoised column * 100)

Oddant1:
Closes #115. Adds a new `percenteage-merged` column caclulated as `merged / denoised * 100` to the output from `denoise_paired`. Currently the 3 tests affected by this are failing, and I am unsure why. I changed the test data files accordingly. When I save the metadata output by `denoise_paired` to disk as a .tsv, then import it into ipython as a dataframe and compare it to the test data file imported as a dataframe, they are equal, but for some reason they are not equal when compared in the actual tests. When I checked the values of the expected and resultant dataframes in the tests, I found they only showed 7 decimals and not the 13 I see when I test them in ipython, but all decimals displayed were identical in spite of the supposed inequality.


Oddant1:
Unsurprisingly, these failures seem to be caused by some floating point shenanigans. I will look into it more tomorrow.

Oddant1:
The commit https://github.com/qiime2/q2-dada2/pull/120/commits/c6e47efaf431d30110ea5915e1316168a31413b0 arbitrarily rounds the float values to the nearest two decimal places which unsurprisingly fixes the floating point shenanigans, I don't know if it's what we want though.

ebolyen:
I think this strategy is fine, but I do wonder if we shouldn't have a nicer visualization for this output at this point. It's a fairly fundamental part of our workflow...

q2_dada2/_denoise.py
```diff
@@ -247,7 +247,12 @@ def denoise_paired(demultiplexed_seqs: SingleLanePerSamplePairedEndFastqDirFmt,
                 raise Exception("An error was encountered while running DADA2"
                                 " in R (return code %d), please inspect stdout"
                                 " and stderr to learn more." % e.returncode)
-        return _denoise_helper(biom_fp, track_fp, hashed_feature_ids)
+        table, rep_sequences, md = _denoise_helper(biom_fp, track_fp,
+                                                   hashed_feature_ids)
+        df = md.to_dataframe()
+        df['percentage-merged'] = [round(m / d * 100, 2) for m, d
```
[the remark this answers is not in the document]
Oddant1: Are you sure that's a good idea given that this stat is only going to show up if they ran `denoise_paired`? If we put the calculation in `_denoise_helper` we still only want to execute it if the call to helper came from paired meaning we have to do some additional checks to determine whether that happened or not while we can currently just do it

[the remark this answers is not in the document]
thermokarst: > Are you sure that's a good idea given that this stat is only going to show up if they ran `denoise_paired`?

I think so! You have the advantage of mutating the dataframe _before_ it is converted to `qiime2.Metadata`, which is nice (moving into the util would eliminate the need to cast and the recast).



> If we put the calculation in `_denoise_helper` we still only want to execute it if the call to helper came from paired meaning we have to do some additional checks to determine whether that happened or not while we can currently just do it

Sounds like a good solution to me, and is going to be less expensive than the df-md-df-md casting cycle.

Another option is for the util to return a df, and leave it to the calling site to cast to md, your call.

q2_dada2/tests/test_denoise.py
```diff
@@ -165,7 +165,6 @@ def test_defaults(self):
         # NOTE: changing the chimera_method parameter doesn't impact the
         # results for this dataset
         table, rep_seqs, md = denoise_paired(self.demux_seqs, 150, 150)
-
```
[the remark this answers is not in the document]
thermokarst: Please revert this no-op change.

q2_dada2/_denoise.py
```diff
@@ -119,6 +125,7 @@ def _denoise_helper(biom_fp, track_fp, hashed_feature_ids):
         rep_sequences = DNAIterator(
             (skbio.DNA(id_, metadata={'id': id_})
              for id_ in table.ids(axis='observation')))
+
```
thermokarst: Please revert this no-op change.

q2_dada2/_denoise.py
```diff
@@ -247,6 +254,7 @@ def denoise_paired(demultiplexed_seqs: SingleLanePerSamplePairedEndFastqDirFmt,
                 raise Exception("An error was encountered while running DADA2"
                                 " in R (return code %d), please inspect stdout"
                                 " and stderr to learn more." % e.returncode)
+
```
thermokarst: Please revert this no-op change.

Oddant1: How do I keep doing that :laughing:

q2_dada2/_denoise.py
```diff
@@ -99,6 +99,12 @@ def _denoise_helper(biom_fp, track_fp, hashed_feature_ids):
     df = pd.read_csv(track_fp, sep='\t', index_col=0)
     df.index.name = 'sample-id'
     df = df.rename(index=_filepath_to_sample)
+
+    if 'merged' in df:
+        percentage_merged = [round(m / d * 100, 2) for m,
```
thermokarst: I think that the more idiomatic pandas approach would involve not iterating over the rows, but rather just performing column manipulations. For example:

```python
x = pd.DataFrame([[1,2,3],[4,5,6]], columns=["foo", "bar", "baz"], index=["s1", "s2"])
x["xyz"] = x["foo"] / x["bar"] # adds new column, "xyz"
x = x[["foo", "bar", "xyz", "baz"]] # reorders columns
```

q2_dada2/_denoise.py
```diff
@@ -99,6 +99,13 @@ def _denoise_helper(biom_fp, track_fp, hashed_feature_ids):
     df = pd.read_csv(track_fp, sep='\t', index_col=0)
     df.index.name = 'sample-id'
     df = df.rename(index=_filepath_to_sample)
+
+    if 'merged' in df:
+        percentage_merged = round(df['merged'] / df['denoised'] * 100, 2)
+        # We want to put percentage-merged right after merged
+        inerstion_index = df.columns.get_loc('merged') + 1
```
thermokarst:
```suggestion
        insertion_loc = df.columns.get_loc('merged') + 1
```

thermokarst: Oops, you still have a typo in the variable name, @Oddant1.

nbokulich:
@Oddant1 do you have an example QZV that I can look at for a UX review?

thermokarst:
this is a QZV (with a zip extension for GH). 
[denoising-stats.zip](https://github.com/qiime2/q2-dada2/files/3643280/denoising-stats.zip)

q2_dada2/_denoise.py
```diff
@@ -99,6 +99,13 @@ def _denoise_helper(biom_fp, track_fp, hashed_feature_ids):
     df = pd.read_csv(track_fp, sep='\t', index_col=0)
     df.index.name = 'sample-id'
     df = df.rename(index=_filepath_to_sample)
+
+    if 'merged' in df:
+        percentage_merged = round(df['merged'] / df['denoised'] * 100, 2)
```
nbokulich: add a line:
```
percentage_merged = percentage_merged.fillna(0)
```

otherwise NaNs render as empty space, which is just visually jarring and may raise questions with obvious answers:
image

q2_dada2/_denoise.py
```diff
@@ -99,6 +99,13 @@ def _denoise_helper(biom_fp, track_fp, hashed_feature_ids):
...
+        # We want to put percentage-merged right after merged
+        insertion_loc = df.columns.get_loc('merged') + 1
+        df.insert(insertion_loc, 'percentage-merged', percentage_merged)
```
nbokulich: Is there any way to make clear that this is merged / denoised? as opposed to merged/input? Maybe call this column `merged / denoised (%)`???

while you are at it, maybe at another column: `non-chimeric / input (%)`

Oddant1: How are we deciding which extra percentage columns we need? Is there any reason not to have percentage columns for every step? Like `percentage_filtered` and `percentage_denoised`

nbokulich: good question. I am picking what _I think_ are the most important for diagnosing run quality, but that is probably a highly subjective criterion! It is probably worth adding a % filtered, but # denoised almost always == # filtered so is probably not too informative and would just crowd the view.

non-chimeric / input would be more useful that non-chimeric / merged because (1) this would apply for single-end and paired-end alike and (2) the point of this is to measure the % of input reads that are output by dada2, rather than the impact of intermediate steps (so a better name may be `output / input %`). For the intermediate steps, it is better (in my opinion) to measure the incremental change, though I suppose if all columns are % of input then it is easy to subtract if needed to assess incremental changes and more intuitive overall.

What do you think @Oddant1 @thermokarst ?

nbokulich: IMO yes, though now I wonder if for the sake of consistency %of-input-merged is better than %of-denoised-merged

(I don't like those naming conventions, though — we should use something more human-ready since these data are really just used for human visualization. If we use a consistent denominator we can just add a note describing what that denominator is, and use a shorter column name, e.g., "% merged" or even just "%")

Oddant1: How and where would we add such a note describing what the denominator is?

nbokulich: oh good point — this is visualized with `tabulate`, this is not generating the QZV... so just ignore the note idea. long column names it is!

Oddant1: I'm not sure about whether it should be merged/denoised or merged/input. We are merging the output of the denoise step just merging the input, so I think I lean more towards merged/denoised, but I do like the idea of them all being consistently a percentage of input. Especially since we don't have a column fully defining the percentage of input that was denoised (we have the percentage of input that was filtered, but that will only usually be the same as the percentage of input that was denoised).

q2_dada2/_denoise.py
```diff
@@ -137,10 +137,12 @@ def _denoise_helper(biom_fp, track_fp, hashed_feature_ids):
 # new_col_loc is the name of the column the new column will be inserted after
 def _add_percentage_column(df, numerator_col, denominator_col,
                            new_col_loc, new_col_name):
+    new_col_df = df
```
thermokarst: This actually doesn't quite work as you would expect:

image

Oddant1: Ah ok

Oddant1: I'm assuming there is a copy utility

thermokarst: BTW, we have _all_ been here, python's shallow vs deep copy is not immediately apparent, until it bites you!

nbokulich: `df = df.copy()`

do you really need to preserve the original though? (sorry not looking at the code, just making assumptions based on previous review)

thermokarst: > do you really need to preserve the original though?

I would prefer that.

thermokarst:
Hey @nbokulich, take a look at this [latest QZV](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fforum.qiime2.org%2Fuploads%2Fshort-url%2F8vEs1WdhK9oUkKnXn1Yh0vaCDMj.qzv). I think the `percentage-of-input-filtered` is confusing, it seems like it would be more clear if the value in that column was `(input - filtered )/ input`. What do you think?

nbokulich:
Works for me!

thermokarst:
Sounds good --- @Oddant1, I have a handful of changes I am working on here, I will ping you once those are ready.

thermokarst:
Okay @nbokulich, :basketball: :+1: 

[QZV](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fforum.qiime2.org%2Fuploads%2Fshort-url%2F4fm1FHejgdSHr50BOsWNtbLqnhQ.qzv)

Oddant1:
Those three separate commits have to do with some weirdness in attempting to merge my code with Matt's commit, but the final commit should contain what we want

Issue #87, opened 2018-02-03
Add pooling options to Q2 workflows


benjjneb:
**Improvement Description**
Add a new option that allows users to pick independent sample processing (as done currently), pooled sample processing, or ["pseudo-pooling" that was added in 1.7.5](http://benjjneb.github.io/dada2/pseudo.html). It probably makes sense to wait until the R package 1.8 release is available (~June) to add this.

The pooling options provide better detection of rare per-sample variants at the cost of increased computation time.

Also consider making pseudo-pooling the default processing mode.

**References**
["pseudo-pooling" that was added in 1.7.5](http://benjjneb.github.io/dada2/pseudo.html)

nbokulich:
[forum xref](https://forum.qiime2.org/t/add-dada2-parameters/8345?u=nicholas_bokulich)

benjjneb:
Question: Can default parameter choices be dependent on other parameter choices?

The reason I ask: Pooled chimera removal is better if pooled sample inference is performed, but the default chimera removal is consensus, which is better for the default sample inference method (independent). So, can chimera removal be defaulted to pooled if the user selects pooled sample inference?

ebolyen:
Not really. There would be a way to refine the types based on other types passed (should be available next week-ish), but that would categorically prevent mixing the two.

A different approach would be to have the two steps be separate actions, and then in a pipeline which composes them, you have a "simpler" argument which unifies the two arguments. That way, the "default" invocation does the ideal thing for inference and chimera checking, but mixing them is still possible if you run the sub-actions directly.

benjjneb:
My current idea is to change the default chimera method to "auto", which chooses "consensus" or "pooled" chimera removal depending on the choice made at the sample inference step. Users will still be able to define the chimera removal method themselves in which case that choice will be used.

That achieves my goal here of defaulting to the "right" chimera removal method for each sample inference method, but let me know if that seems a bad idea.

ebolyen:
That works too! There are a few places where we have similar patterns.

yanxianl:
Hi, will pseudo-pooling and pooling make it to the coming release of QIIME2-2019.7? I'm looking forward to using pseudo-pooling for my dataset within QIIME2, which otherwise has to be done in R instead.

benjjneb:
Shoot, I didn't have this Q2 release on my calendar and it looks like PRs
are due July 22.
It would be really nice to get pseudo-pooling in though, as I know there
[...]

PR #122, opened 2019-10-28
Pseudo pool

benjjneb:
This pull request adds pseudo-pooling to all DADA2 workflows: denoise-single, denoise-pyro, and denoise-paired. It is implemented in a way that should keep memory costs flat, and barely higher than they are when not using pseudo-pooling.

Pseudo-pooling is activated by the new parameter `pooling-method = pseudo`. It's main effect is to increase sensitivity to rare variants in samples, particularly singletons, as long as they were observed in other samples independently.

This passes testing on my end based on Rscript invocations, but python based testing still hasn't been performed, and no new pseudo-pooling specific tests have been added.

This addresses https://github.com/qiime2/q2-dada2/issues/87 but I haven't added full pooling yet (which is much more memory/time intensive).

---

I'm not sure how to deal with the unit-tests and python testing. In principle, the current tests should all pass as the output using the default `pooling-method = independent` should be unchanged from the previous output.

I realize this is probably too late for the upcoming release.


ebolyen:
Working on this today, @benjjneb sorry for the excessive delay.

q2_dada2/tests/data/expected/pyro-default.tsv
```diff
@@ -1,34 +1,37 @@
...
-707f5ba87a63b8071f5b7cad27aa9046	0.0	0.0	0.0	0.0	3.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0
+5179f5cac20c1aa00377e4942bd7f837	0.0	2.0	3.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0
+707f5ba87a63b8071f5b7cad27aa9046	0.0	0.0	0.0	0.0	3.0	0.0	0.0	0.0	2.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0
 ef9ecba707ad092c6ac7bef17d0db268	0.0	0.0	0.0	0.0	0.0	0.0	0.0	3.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0
-a7f92909575fd948fad074b72093771e	0.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0	3.0	0.0	0.0	0.0
\ No newline at end of file
+a7f92909575fd948fad074b72093771e	0.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0	3.0	0.0	0.0	0.0
+9d07a6ff18a64d9e41e70bb17e9e02c7	0.0	0.0	2.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0
+857b4205ec3f91149633495d70354521	0.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0	2.0	0.0	0.0	0.0	0.0	0.0	0.0
+4629fbb5dde5f8ee07c2259d2887a8d4	0.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0	0.0	2.0
```
ebolyen: I wasn't really expecting the pyro tests to change, but this seems reasonable. @benjjneb does this look alright? (I sorted the output to match the old one so that the diff could be checked).

benjjneb: Those don't look like major shifts, but I'm also not expecting there to be any change in the pyro results. 

Is my understanding correct that only the denoise-pyro results changed, and that the denoise-single and denoise-paired tests remained identical?

ebolyen: That is correct, the defaults on single and paired did not change, only pyro.

q2_dada2/assets/run_dada_paired.R
```diff
@@ -199,15 +209,51 @@ errR <- suppressWarnings(learnErrors(filtsR, nreads=nreads.learn, multithread=mu
...
+  drpF <- derepFastq(filtsF[[j]])
+  ddsF[[j]] <- dada(drpF, err=errF, multithread=multithread, verbose=FALSE)
+  drpR <- derepFastq(filtsR[[j]])
+  ddsR[[j]] <- dada(drpR, err=errR, multithread=multithread, verbose=FALSE)
+  cat(".")
+}
+cat("\n")
+if(poolMethod == "pseudo") {
+  cat("  Pseudo-pool step ")
+  ### TEMPORARY, to be removed once 1.12 makes its way to Q2
```
ebolyen: Incidentally, it looks like bioconda has 1.14 available at this point, we could update the recipe to pin at that version if it was helpful.

q2_dada2/tests/data/expected/pyro-default.tsv
benjjneb: Hmm... OK I'll run the pyro tests locally and try to identify what exactly is going on with pyro. Unexpected changes in output are no bueno, but it's probably something obvious I'm missing.

q2_dada2/assets/run_dada_paired.R
benjjneb: 1.14 will allow `denoise-pacbio` to be added, but I wouldn't want to bump the version before I sort out the `denoise-pyro` discrepancy above.

thermokarst:
Hey @benjjneb - no rush to reply, but are there any updates on the denoise-pyro discrepancies? We are about 2 weeks out from the next scheduled QIIME 2 release, and trying to see what outstanding PRs might make the release. Do you think we should include this PR, or defer to the next release (tentatively August)?

thermokarst:
Great, thanks for the update, @benjjneb. Let me know what we can do on our end, when you know it, and we'll be there.

q2_dada2/tests/data/expected/pyro-default.tsv
benjjneb: Should be fixed in e91ee9403195c92270fd9635ee7a711a8d19ea1c

There was indeed a misindexing that had cropped up!

benjjneb:
The `denoise-pyro` bug that caused the change in test results should be fixed by e91ee9403195c92270fd9635ee7a711a8d19ea1c

PR #127, opened 2020-01-17
BUG: fix single sample recognized as empty table

ebolyen:
This was leftover in my working tree, I couldn't find the issue, but the change made sense, so I've finished the implementation with some tests.


q2_dada2/tests/data/single_sample.tsv
```diff
@@ -0,0 +1,2 @@
+#OTU ID	L1S208	L1S257	L1S57	L1S76	L2S155	L2S175	L2S309	L2S357	L3S294	L3S313	L4S112	L4S63	L5S155	L5S174	L6S20	L6S68
```
thermokarst: Are we testing a single sample (`single_sample.tsv`) or a single feature (`single_feature.tsv`)? If its the first can you explain in more detail? Thanks!

ebolyen: good catch, that should be `single_feature` I'll rename this file.

q2_dada2/tests/test_denoise.py
```diff
@@ -292,5 +293,23 @@ def test_trunc_len_bigger_than_max_len(self):
         denoise_pyro(self.demux_seqs, 100, max_len=160)
 
 
+class TestUtils(TestPluginBase):
+    package = 'q2_dada2.tests'
+
+    def test_check_featureless_table_single_sample(self):
```
ebolyen: same issue here

q2_dada2/tests/data/single_sample.tsv
thermokarst: Please rename the tests etc as [necessary](https://github.com/qiime2/q2-dada2/pull/127#pullrequestreview-350268631), too!

thermokarst:
Tested locally, gonna merge before travis finishes.

Issue #101, opened 2018-09-20
Add long-read support


benjjneb:
**Improvement Description**
Add long-read support

**Current Behavior**
We've added support for PacBio long-read amplicon sequencing to the devel version of the R package and it seems to work quite well. Preprint: High-throughput amplicon sequencing of the full-length 16S rRNA gene with single-nucleotide resolution.

**Proposed Behavior**
I think it will make sense to add this as a tech-specific ```denoise-pacbio``` command in the plugin. This is similar to the ```denoise-pyro``` approach already in the plugin, with the purpose of having a dedicated command being to automatically turn on the right flags and options for PacBio data rather than relying on the user to do so. There is a downside in the repetition of much of the code between the different ```denoise-[technology]``` commands.

**References**
1. Preprint: [High-throughput amplicon sequencing of the full-length 16S rRNA gene with single-nucleotide resolution](https://www.biorxiv.org/content/early/2018/08/15/392332).

benjjneb:
This won't be in the 2019Q2 update. Tentatively targeting this for end-of-year 2019.

colinbrislawn:
What's the current status? People love those long reads! 🧬

benjjneb:
Waiting on the R package 1.12 to propagate through bioconda to Q2, as that made some important long read improvements.

benjjneb:
IMO, this is the next bit of functionality I'd really like to propagate through to the plugin.

First step would be to upgrade Q2 to the 1.14 version of the package or later. Then, it will be pretty straightforward to add a `denoise-pacbio` command that basically uses `denoise_single`/`_denoise_single` just like `denoise-pyro` does to avoid much repetition. 

Also consider adding `denoise-loop` in similar fashion (for Loop Genomics tech).

Oddant1:
Are we ready to try to do this? It looks like we're still on `bioconductor-dada2` version 1.10.0

benjjneb:
Can't add this functionality until Q2's version `bioconductor-dada2` is upgraded to 1.14 or later.

Versions up to 1.16 are already available through bioconda: https://bioconda.github.io/recipes/bioconductor-dada2/README.html

harish0201:
I finished installing qiime2 recently, and DADA2 is still 1.10.

When are the upgrades planned to incorporate PacBio Long Reads? 

@benjjneb Would you suggest upgrading the installed version and modifying the existing script (for Pyro or single ends) to incorporate PacBio specific error profiles?

[...]

PR #135, opened 2021-01-30
add support for Pacbio CCS reads

sixvable:
Add support for Pacbio CCS reads.
issue #101 . 
Test on **qiime2-2020.11**.

### DIFF
--- a/q2_dada2/_denoise.py
+++ b/q2_dada2/_denoise.py
@@ -61,10 +61,12 @@ _valid_inputs = {
     'trim_left': _WHOLE_NUM,
     'trim_left_f': _WHOLE_NUM,
     'trim_left_r': _WHOLE_NUM,
+    'max_mismatch': _WHOLE_NUM,
     'max_ee': _NAT_NUM,
     'max_ee_f': _NAT_NUM,
     'max_ee_r': _NAT_NUM,
     'trunc_q': _WHOLE_NUM,
+    'min_len': _WHOLE_NUM,
     'max_len': _WHOLE_NUM,
     'pooling_method': _POOL_STR,
     'chimera_method': _CHIM_STR,
@@ -78,6 +80,9 @@ _valid_inputs = {
     'demultiplexed_seqs': _SKIP,
     'homopolymer_gap_penalty': _SKIP,
     'band_size': _SKIP,
+    'front': _SKIP,
+    'adapter': _SKIP,
+    'indels': _SKIP,
 }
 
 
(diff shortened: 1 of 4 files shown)


benjjneb:
Thanks for working on this @sixvable 

I skimmed the code and two questions come to mind immediately.

1) Are the fraction of reads that pass primer removal included in the final tracking data.frame that can be reported out to the Q2 user? This is important information.

2) This is an entirely new script, but one that heavily resembles `denoised-single`. Long-term, does it make more sense to make a new script like this, or to leverage `denoise-single` with the differences fed in as options to that script?

sixvable:
Thank you @benjjneb 



> 1. Are the fraction of reads that pass primer removal included in the final tracking data.frame that can be reported out to the Q2 user? This is important information.

Yes, the number and percentage of the reads that pass the primer remove step have been include in the track file, and can be check in the final denoising_stats.qzv


> 2\. This is an entirely new script, but one that heavily resembles `denoised-single`. Long-term, does it make more sense to make a new script like this, or to leverage `denoise-single` with the differences fed in as options to that script?

Since there is a new *primer remove* step before the *trim and filter* step in the `denoise-ccs` function, I had to re-arrange the sort numbers of the entire `run_dada_single.R` script. I also exposed 3 new parameters in the new `run_dada_ccs.R` script: **maxMismatch**, **indels**, and **minLen**. And I also change the default value of **minQ**. If we directly put this part into `run_dada_single.R`, we may need to give more **ifelse** code and the script could also become less readable or maintainable.

benjjneb:
> Yes, the number and percentage of the reads that pass the primer remove step have been include in the track file, and can be check in the final denoising_stats.qzv

Nice!

> Since there is a new primer remove step before the trim and filter step in the denoise-ccs function, I had to re-arrange the sort numbers of the entire run_dada_single.R script. I also exposed 3 new parameters in the new run_dada_ccs.R script: maxMismatch, indels, and minLen. And I also change the default value of minQ. If we directly put this part into run_dada_single.R, we may need to give more ifelse code and the script could also become less readable or maintainable.

That seems reasonable. I just wanted to bring this issue up, as longer term maintainability is a major issue (although I'm sure that's not news to the Q2 team).

benjjneb:
Another Q: Have you identified a PacBio CCS dataset for the integrated testing?

sixvable:
> Another Q: Have you identified a PacBio CCS dataset for the integrated testing?

There are not many benchmarks on PacBio CCS dataset using DADA2 so I only test several datasets in [your paper](https://github.com/benjjneb/LRASManuscript). The q2 version gives the exactly same results compared to the R version. So I think it is worked.

benjjneb:
That seems like good confirmation it is working as expected. For the integrated testing, it might be best to downsample a single lower-diversity sample (maybe the Zymo mock) just to keep computational expenditures lower.

sixvable:
Zymo mock works well too. I will leave it to q2 team for further test.

q2_dada2/_denoise.py
```diff
@@ -147,6 +152,60 @@ def _denoise_helper(biom_fp, track_fp, hashed_feature_ids):
              for id_ in table.ids(axis='observation')))
     return table, rep_sequences, metadata
 
+def _denoise_helper2(biom_fp, track_fp, hashed_feature_ids):
```
thermokarst, line 155: This utility function is largely a duplicate of the original `_denoise_helper`, given that there is a lot of moving parts wrapped up in here, perhaps it would be safer to merge the two helpers together (at least then we won't forget to update both helpers, when making edits). It seems like most/all of the differences are related to the denoising stats code - perhaps _that_ should be factored out into its own set of utility functions?

sixvable: Yes, you are right. I have refactored ` _denoise_helper2` into the original  `_denoise_helper` function. Sorry for the mistakes.

ebolyen:
@Keegan-Evans, looks like you are missing an entry in `package_data` in `setup.py` to include your test data directory.

Keegan-Evans:
Basic tests and updating to current q2-dada2 version done. Data for testing is 100 sequence subsamples from the human fecal replicate 1 samples from @benjjneb's paper.

Keegan-Evans (approved):
@ebolyen for final review.

ebolyen (approved):
This all looks great to me. Thank you @sixvable for doing the heavy lifting and thank @Keegan-Evans for adding some test data. Sorry this kind of took a while, but it's really exciting to have this!

Issue #136, opened 2021-03-30
dada2 drops samples without reads unfiltered reads from table but not denoising stats.


dwthomas:
**Bug Description**
dada2-paired appears to drop samples which have 0 sequences after filtering, or perhaps denoising, but keeps samples which have zero sequences from merging and beyond.  
This leads to samples being dropped from the table, though they remain in the denoising stats.  

I've not checked the behavior with the dada2 paired or dada2 pyro or with older versions of dada2.

**Steps to reproduce the behavior**
Run a dataset including samples with no reads that pass the filtering step, I had a sample with 0 reads and a sample with 6 poor quality reads that were both dropped.

**Expected behavior**
The samples should be retained with 0 frequency in the table.  

**Computation Environment**
- OS: Linux
- QIIME 2 Release 2021.2

**Questions**
1. Is this a qiime2 issue or a dada2 issue?  

**Comments**
1. Ideally blanks should be blank, so discarding them in dada2 means the user has to actively notice that they are missing.
2. A trivial fix is something like: 
`
for i in set(denoising_stats.index) - set(table.index):
    table.loc[i] = 0
`
Though there may be a more clear way of getting the dropped samples.
I can have a go at implementing that, but I wanted options on if this is something to change in dada2, or here?

Here is the data I noticed this behavior in: https://unh.box.com/s/pyz250peix6hiyrcye2fnwycidx171sz

benjjneb:
> Is this a qiime2 issue or a dada2 issue?

It's sort of a dada2 issue, in that we haven't robustly implemented processing zero-read samples through the whole pipeline, therefore they need to be removed from the processing before entering the post-filtering denoising workflow.

That said, it is probably an easy fix on the Q2 side by just adding those zero-read samples back into whatever final table where they are still wanted, as it seems like you are suggesting.

PR #139, opened 2021-04-02
implemented fix to dada2 dropping blank samples. 

dwthomas:
Fixes issue #136


thermokarst (commented):
Thanks @dwthomas - this sounds like a good approach to me. Two points:

1. Should this behavior be put behind a parameter flag, like `--p-retain-all-samples`? If so, maybe we should make the existing behavior the default setting for that parameter, that way folks can opt _into_ this sample retention scheme. I am concerned that there might be downstream implications of retaining empty samples, and while the solution is simple (filter the empty samples out), it might be a stumbling block for new users. Thoughts?
2. This is missing unit tests. If you want a hand with writing some, just let me know, I am happy to lend a hand.

We are gearing up for the 2021.4 release, so if you want this to ship in this release, we only have another day to make that happen (sorry for the late response here) - just let me know how you want to proceed. Thanks!

dwthomas:
Thanks @thermokarst 
1. As it stands dada2 will retain empty samples if they are not empty until the merging step of dada2, which to me seems like a bug, as you'd want all the empty samples or none.  I think a flag might be a good but separate idea as I agree there are probably some folks assuming there are no empty samples, as you say.
2. Some help navigating the unit tests would be useful!  I was having trouble finding where to put/get the input for a unit test.  I believe just adding another sample with no reads should demonstrate the behavior.

thermokarst:
Thanks @dwthomas. So to make sure I'm following, would an alternative solution be to instead _filter_ all empty samples from the final feature table? If so, why didn't you take that approach here?

dwthomas:
I think the two behaviors that make sense are:
1. All samples input to dada2 are retained, even with zero frequency.
2. Samples with zero frequency after dada2 are removed.
The current behavior retains some but not all zero frequency samples, which I think is a potential source of confusion.
I implemented the first because that's the behavior I want.  But my argument for that behavior as the default is that you can go from that to retaining only non-zero samples with a qiime feature-table filter-samples command, but I do not know a way to add back removed samples without using the Artifact API.

thermokarst:
Awesome, thanks for explaining, that makes sense to me! Okay yeah, I'm leaning towards going with the functionality proposed here. In the future I think it would be great to add in a new flag that lets us toggle the empty-sample retention, wholesale, but I agree, that is a separate issue. I'll think about tests and get back to you. Would you mind if I push the tests up to your feature branch?

dwthomas:
That works great for me.

dwthomas:
@thermokarst I'm doing some work on the unit tests, I figured out where the data files are.

dwthomas:
Hey @thermokarst I added unit tests and a toggle as discussed.  All the functionality is in the _denoise_helper function used by all single, paired and pyro so I modified the paired tests to include an empty sample and test both modes of the retain-all-samples toggle.

q2_dada2/_denoise.py
```diff
@@ -132,6 +135,15 @@ def _denoise_helper(biom_fp, track_fp, hashed_feature_ids):
...
                for id_ in table.ids(axis='sample')}
     table.update_ids(sid_map, axis='sample', inplace=True)
+    # Reintroduce empty samples dropped by dada2.
+    table_cols = table.ids(axis='observation')
+    table_rows = list(set(df.index) - set(table.ids()))
+    table_to_add = biom.Table(np.zeros((len(table_cols), len(table_rows))),
+                              table_cols, table_rows,
+                              type="OTU table")
+    table = table.concat(table_to_add)
+    if not retain_all_samples:
```
thermokarst, line 145: Probably best to comment this to explain that there are other sources of empty samples, not just the 6ish lines above this (otherwise, you might be tempted to stash everything in an if/else statement).

q2_dada2/tests/data/sample_seqs_paired/MANIFEST
```diff
@@ -1,21 +1,23 @@
...
-BAQ2420.2,BAQ2420.2_36_L001_R2_001.fastq.gz,reverse
-YUN1242.3,YUN1242.3_44_L001_R1_001.fastq.gz,forward
-YUN1242.3,YUN1242.3_44_L001_R2_001.fastq.gz,reverse
-BAQ4166.3,BAQ4166.3_48_L001_R1_001.fastq.gz,forward
-BAQ4166.3,BAQ4166.3_48_L001_R2_001.fastq.gz,reverse
-YUN1005.3,YUN1005.3_49_L001_R1_001.fastq.gz,forward
-YUN1005.3,YUN1005.3_49_L001_R2_001.fastq.gz,reverse
-YUN3153.3,YUN3153.3_60_L001_R1_001.fastq.gz,forward
-YUN3153.3,YUN3153.3_60_L001_R2_001.fastq.gz,reverse
+BAQ3473.2,BAQ3473.2_0_L001_R1_001.fastq.gz,forward
```
thermokarst, line 2: What happened with all of these files? It looks like maybe they all got renamed or something?

q2_dada2/tests/test_denoise.py
```diff
@@ -24,6 +24,10 @@ def _sort_seqs(seqs):
     return sorted(list(seqs), key=lambda x: x.metadata['id'])
 
 
+def _sort_table(table):
```
thermokarst, line 27: That wasn't necessary before, and I'm not sure what/why the changes in this PR prompted this change - did you run into a problem with unstable data during these tests?

thermokarst (changes requested):

q2_dada2/_denoise.py
```diff
@@ -54,6 +55,7 @@ def _check_featureless_table(fp):
              'pooled, consensus or none')
 # Better to choose to skip, than to implicitly ignore things that KeyError
 _SKIP = (lambda x: True, '')
+_BOOL = (lambda x: x in {True, False}, "Boolean")
```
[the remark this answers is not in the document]
dwthomas: Great Point!

q2_dada2/_denoise.py
```diff
@@ -132,6 +135,15 @@ def _denoise_helper(biom_fp, track_fp, hashed_feature_ids):
     sid_map = {id_: _filepath_to_sample(id_)
                for id_ in table.ids(axis='sample')}
     table.update_ids(sid_map, axis='sample', inplace=True)
+    # Reintroduce empty samples dropped by dada2.
+    table_cols = table.ids(axis='observation')
+    table_rows = list(set(df.index) - set(table.ids()))
```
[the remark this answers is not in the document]
dwthomas: I get "TypeError: iteration over a 0-d array" when I pass a set to the Biom table constructor in the following line.  Which is why I did the conversion to list.  I agree it seems silly.

q2_dada2/tests/data/sample_seqs_paired/MANIFEST
dwthomas: Yeah they did, I tried adding the "Empty" sample by modifying the Manifest file, but they were still being ignored.  So I imported them into qiime with 2021.2, then exported them, which worked but it seems to have renamed a lot of the files.

q2_dada2/_denoise.py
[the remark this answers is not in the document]
thermokarst: Oh sorry, I missed that below, I thought you were _only_ using this to get the total id count, my mistake!

q2_dada2/tests/test_denoise.py
dwthomas: Biom.table.concat sorts the opposite axis, see line 3391 here: https://github.com/biocore/biom-format/blob/master/biom/table.py, seems weird that it is not an option since it means there is not an equivalent to the additive identity of zero since even if there is nothing to concatenate you sort the opposite axis.

I chose to add the sorting prior to comparing:
1. I wanted to minimize changes in the _denoise code.
2. I did not want to modify the expected tables except adding a column for the new BLANK sample (to make it more clear)
3. This was easy to implement and I think clear what it is doing, maybe I should add a comment about why it became necessary?

q2_dada2/_denoise.py
dwthomas: Good idea!  I inserted one.

thermokarst (approved):

q2_dada2/_denoise.py
```diff
@@ -83,6 +85,7 @@ def _check_featureless_table(fp):
     'demultiplexed_seqs': _SKIP,
     'homopolymer_gap_penalty': _SKIP,
     'band_size': _SKIP,
+    "retain_all_samples": _BOOL
```
thermokarst, line 88:
```suggestion
    'retain_all_samples': _BOOL,
```

q2_dada2/_denoise.py
```diff
@@ -419,4 +443,4 @@ def denoise_ccs(demultiplexed_seqs: SingleLanePerSampleSingleEndFastqDirFmt,
                 raise Exception("An error was encountered while running DADA2"
                                 " in R (return code %d), please inspect stdout"
                                 " and stderr to learn more." % e.returncode)
-        return _denoise_helper(biom_fp, track_fp, hashed_feature_ids)
+        return _denoise_helper(biom_fp, track_fp, hashed_feature_ids)
```
thermokarst, line 446:
```suggestion
        return _denoise_helper(biom_fp, track_fp, hashed_feature_ids)

```

dwthomas:
Looks like there is just a new dada2 flavor that needed the toggle added to it as well.

Oddant1:
Will this be ready to merge after the conflict is resolved?

Oddant1:
@dwthomas are you still working on this PR?

cherman2:
Hi @dwthomas! 
We are currently cleaning out our forum backlog! I fixed conflicts but it looks like there are still some errors! Do you have the bandwidth to take this pull request across the finish line? No pressure if not!

dwthomas:
Hey @cherman2,
I unfortunately don't have the bandwidth at least for the time being.

cherman2:
No problem! 
All these remaining issues look super superficial  (and I may have caused them when I fixed conflicts). I will get this passing and then pull down to test locally!

cherman2 (approved):
This looks good to me as long as CI passes! 👏 

@dwthomas, Thank you for all your work on this! This is awesome.

hagenjp (approved):
looks good to me! Thank you, @dwthomas

PR #141, opened 2021-06-08
dep/pin: r-matrix 1.3-2

thermokarst:


benjjneb:
Sorry I should have mentioned this issue, which was the subject of some consternation on the DADA2 issues forum: https://github.com/benjjneb/dada2/issues/212#issuecomment-360994743

Pinning Matrix to 1.3-2 should be a good fix for nwo.

Longer term, all internal DADA2 usage of the offending `ShortRead` package code was removed in our latest release, so this will be a dead issue once that version (1.20) propagates to Q2.

PR #145, opened 2021-12-01
DEP: dada2 1.22.0

Keegan-Evans:


thermokarst:
We tested this out locally using a minimal q2 env:

```bash
conda create \
  -n dada2 \
  -c https://packages.qiime2.org/qiime2/2022.2/tested/ \
  -c conda-forge \
  -c bioconda \
  -c defaults \
  qiime2 q2cli q2-types bioconductor-dada2==1.22.0
```

Unless we messed something up, none of the existing unit tests failed (which was surprising to me, I guess just based on historical precedent, it seems like the unit tests here usually need some massaging when bumping dada2 versions).

PR #146, opened 2022-03-10
Support allowOneOff option in removeBineraDenovo()

### LINKED ISSUE #126: Setting "allowOneOf" and "maxMismatch" parameters in q2/dada2 plugin
Hello,

We were used to work with dada2 1.10 and now we want to move to q2 (2019.10) using the DADA2 plugin embedded in q2.

By reading carefully? the documentation of the dada2 plugin (https://docs.qiime2.org/2019.10/plugins/available/dada2/denoise-paired/) we did not find how to adjust the following parameters :

-	allowOneOff of the DADA2 removeBimeraDenovo method (following this discussion: https://github.com/benjjneb/dada2/issues/610)
-	maxMismatch in the DADA2 mergePairs method (we prefer this over having a fixed quality score / sequence length cutoff)

Is there any way to edit these parameters in q2-dada2 ?

Thanks a lot

Francis

leasi:
Support of the `allowOneOff` argument in the `runBimeraDenovo()` function, which enables the user to allow for one mismatch or indel in bimera detection when set to `True` (default = `False`).
An example of the impact of this parameter is described in benjjneb/dada2#610.
Resolves #126.

Took the opportunity to correct the parameters description & order and command example in the header of the R scripts.

(Note: this is my first pull request ever so hopefully I did everything properly😊 )

### DIFF
--- a/q2_dada2/assets/run_dada_single.R
+++ b/q2_dada2/assets/run_dada_single.R
@@ -6,7 +6,7 @@
 # table. It is intended for use with the QIIME2 plugin
 # for DADA2.
 #
-# Ex: Rscript run_dada_single.R input_dir output.tsv track.tsv filtered_dir 200 0 2.0 2 Inf pooled 1.0 0 1000000 NULL 32
+# Ex: Rscript run_dada_single.R input_dir output.tsv track.tsv filtered_dir 200 0 2.0 2 Inf independent pooled 1.0 TRUE 0 1000000 NULL 32
 ####################################################
 
 ####################################################
@@ -79,6 +79,12 @@
 #               abundant than the sequence being tested).
 #    Ex: 1.0
 #
+# 13) allowOneOff - Bimeras that are one-off from exact are also identified if the allowOneOff
+#               argument is TRUE. If FALSE, a sequence will be identified as bimera if it is one 
+#               mismatch or indel away from an exact bimera.
+#    Ex: FALSE
+#
+#
 ### SPEED ARGUMENTS ###
 #
 # 13) nthreads - The number of threads to use.
@@ -88,6 +94,7 @@
 # 14) nreads_learn - The minimum number of reads to learn the error model from.
 #                 Special values: 0 - Use all input reads.
 #    Ex: 1000000
+#
 #
 ### GLOBAL OPTION ARGUMENTS ###
 #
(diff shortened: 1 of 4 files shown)


ebolyen:
Hi @leasi!

Thank you for this PR and great job on your first one, I've just approved CI to run and it looks like there's some minor lint errors. You should be able to view the log for more information on that.

Overall this all looks in order, but I do want to take a closer look tomorrow, since there's so many parameters. Thank you for updating the "documentation" in the R script as well. It looks like that got out of sync at some point :)

leasi:
Hello @ebolyen, thank you for the quick and lovely feedback! I corrected the formating errors so hopefully lint will no longer complain. Let me know if there's anything else I can do to help!

Keegan-Evans:
@leasi, it looks like there are some test failures, but @lizgehret and I are going to look into it in the morning!

q2_dada2/assets/run_dada_single.R
```diff
@@ -119,12 +126,13 @@ maxLen <- as.numeric(args[[9]]) # Allows Inf
...
-nthreads <- as.integer(args[[13]])
-nreads.learn <- as.integer(args[[14]])
+allowOneOff <- as.logical(args[[13]])
+nthreads <- as.integer(args[[14]])
+nreads.learn <- as.integer(args[[15]])
 # The following args are not directly exposed to end users in q2-dada2,
 # but rather indirectly, via the methods `denoise-single` and `denoise-pyro`.
-HOMOPOLYMER_GAP_PENALTY <- if (args[[15]]=='NULL') NULL else as.integer(args[[15]])
-BAND_SIZE <- as.integer(args[[16]])
+HOMOPOLYMER_GAP_PENALTY <- if (args[[16]]=='NULL') NULL else as.integer(args[[15]])
```
lizgehret, line 134:
```suggestion
HOMOPOLYMER_GAP_PENALTY <- if (args[[16]]=='NULL') NULL else as.integer(args[[16]])
```

lizgehret: @leasi looks like you just missed adjusting this index - with this change, all existing tests within q2-dada2 pass! 🥳

lizgehret (changes requested):
One small adjustment and then you should be good to go @leasi!

leasi:
Definitely an overlook on my end, thanks for the quick fix @lizgehret!

Keegan-Evans (approved):

Keegan-Evans:
An easy thing to overlook! If @lizgehret can approve in the "change requested" has been resolved, I think @ebolyen can get this merged next week when he gets back to the office. Thanks @leasi!

ebolyen:
This is great, thanks everyone!

I have one more minor comment/clarification on the language of the help text (this exists in DADA2's documentation as well).

I read your thread and it sounds like `allowOneOff` changed from TRUE to FALSE, resulting in more "suspected bimeras" making it through the filtering. It also sounds like the new default should indeed be false.

So I take it that the "allow" is in reference to the selection of a bimera. i.e. an ASV which is an indel or mismatch from another ASV would be marked as chimeric if `allowOneOff=TRUE`. This doesn't really match what the documentation says.

Looking at the history for DADA2, it looks like the documentation was incorrectly updated, as my understanding is the goal was to change the default behavior (not the argument which invokes the default behavior): https://github.com/benjjneb/dada2/commit/3750df1940b6b206f0ee96a6fcd0daa667c4bb90

Assuming my interpretation is correct, can we update the text here, as it seems to contradict itself.

cc @benjjneb

---

Unrelated feature, if you name a parameter with backticks: ``` `parameter_name` ``` (in this case ``` `allow_one_off` ```) q2cli should recognize it in the text and underline it for you. That's maybe more useful when referencing a different parameter.

benjjneb:
Thanks @ebolyen you are correct, that was a mistake in the DADA2 help text, for exactly the reason you diagnosed!

`allowOneOff` should be `FALSE` by default, and when `TRUE` causes bimeras that are one-off of a perfect. bimeric model to be identified and removed.

leasi:
The devil is in the details, thanks for catching this @ebolyen, and for the confirmation @benjjneb (and for already opening the appropriate benjjneb/dada2#1518 issue in the dada2 original repo)!

I corrected the help accordingly in this PR. Took the opportunity to properly describe the parameter in the help using backticks as suggested!

PR #152, opened 2022-10-24
ex: adds usage examples for Hacktoberfest

lizgehret:
This PR is part of our 2022 Hacktoberfest team-wide sprint to add usage examples to all methods in each of our core distribution plugins. The methods examples have been added for are as follows:

```
denoise single
denoise paired
```

### DIFF
--- a/q2_dada2/tests/test_denoise.py
+++ b/q2_dada2/tests/test_denoise.py
@@ -24,6 +24,13 @@ def _sort_seqs(seqs):
     return sorted(list(seqs), key=lambda x: x.metadata['id'])
 
 
+class TestBase(TestPluginBase):
+    package = 'q2_dada2.tests'
+
+    def test_examples(self):
+        self.execute_examples()
+
+
 class TestDenoiseSingle(TestPluginBase):
     package = 'q2_dada2.tests'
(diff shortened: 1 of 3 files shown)


q2_dada2/tests/test_denoise.py
```diff
@@ -24,6 +24,13 @@ def _sort_seqs(seqs):
     return sorted(list(seqs), key=lambda x: x.metadata['id'])
 
 
+class TestBase(TestPluginBase):
```
ChrisKeefe, line 27: Is TestBase the best name for this?

lizgehret: Good question @ChrisKeefe - I pulled this from the formatting of our other unit tests (I've put usage examples in there before along with the testing set up) for future usability if any factories need to be initialized there, but I'm not attached to it - I could change it to TestExamples instead. Or add the test_examples method under the TestUtils() class?

lizgehret: For posterity - we discussed offline and I have updated this class to TestExamples()

Oddant1 (approved):
Looks good to me

PR #153, opened 2022-11-07
imp: adding R traceback for better error handling

lizgehret:
This PR adds a preliminary code block that will provide an R traceback for errors encountered within this R wrapper. I've also done some minor linting in this file to remove extra whitespace.

@lina-kim I wanted to share this with you - @nbokulich had mentioned this could be something useful for the work you're doing in R, so you may want to add this to any of your existing R scripts if that's helpful for you!

### DIFF
--- a/q2_dada2/assets/run_dada.R
+++ b/q2_dada2/assets/run_dada.R
@@ -128,6 +128,16 @@
 #    Ex: 32
 #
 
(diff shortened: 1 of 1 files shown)


Oddant1 (approved):
I don't really know R, but that looks pretty good to me.

PR #154, opened 2022-11-15
dep: adding r-base

lizgehret:
this PR min pins r-base to keep dada2 up to date with the global r-base requirement in package integration.

### DIFF
--- a/ci/recipe/meta.yaml
+++ b/ci/recipe/meta.yaml
@@ -19,8 +19,7 @@ requirements:
   run:
     - python {{ python }}
     - biom-format {{ biom_format }}
-    - bioconductor-dada2 1.22.0
-    - r-base >=4.2.0
+    - r-base {{ r-base }}
     - r-optparse >=1.7.1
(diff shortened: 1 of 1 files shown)


ci/recipe/meta.yaml
```diff
@@ -20,6 +20,7 @@ requirements:
     - python {{ python }}
     - biom-format {{ biom_format }}
     - bioconductor-dada2 1.22.0
+    - r-base >=4.2.0
```
ebolyen, line 23:
```suggestion
    - r-base {{ r_base }}
```

We should set this via `conda_build_config` so that we don't have to update this everywhere each time there is a new bioconductor

ebolyen (changes requested):
Would it be worth just dropping the version on `bioconductor-dada2`? The new `r-base` pin should keep us within 1 or 2 bioconductor releases depending on their release speed vs R.

ebolyen (approved):
Tests looks good and the latest versions were pulled!

Issue #163, opened 2023-12-11
BUG:Can not run qiime dada2 denoise-ccs without primers or adapters 


cherman2:
**Improvement Description**
We need to be able to run qiime dada2 denoise-ccs even if the data does not have primers or adapters. some users are receiving their data from sequencers without adapters, primers, and the sequences and merged 

**Current Behavior**
qiime dada2 denoise-ccs requires   --p-front and --p-adapter are required fields. Looking at the code, I believe this is something we are enforcing on the qiime2 interface side and it does not seem to be a requirement of dada2.

**Proposed Behavior**
Users on the forum have said that they are receiving data without primers and adapters and therefore can not run denoise-ccs

**Comments**
1. users does note that many sequences are not being classified because the sequences are reverse compliment. 

**References**
1. [ Forum X-Ref](https://forum.qiime2.org/t/qiime-dada2-denoise-ccs-without-primers-or-adapters/28546)

lizgehret:
I haven't looked at this in detail, but offhand this seems like a good one for you to work on @hagenjp - this should just be modifying the `--p-front` and `--p-adapter` fields to not be required on the QIIME 2 side. Assigning this to you in case you want a break from all of your migration work, but no immediate rush on this!

yongmingwang12syp:
I had the same problem, but what puzzled me was why someone could run the code without writing those two parameters.

lida56:
Hi, 
I am the first time user of Qiime and found this conversation as I too have problem using "qiime dada2 denoise-ccs". Here is the script I used 
qiime dada2 denoise-ccs --i-demultiplexed-seqs ${dir}/demultiplexed-trimmed-single.qza --p-trim-left 20 --p-trunc-len 1600 --p-n-threads 20 --o-representative-sequences ${dir}/rep-seqs-dada2.qza --o-table ${dir}/table-dada-test1.qza --o-denoising-stats ${dir}/stats-dada2-test1.qza --p-front AGRGTTYGATYMTGGCTCAG --p-adapter AAGTCGTAACAAGGTARCY --verbose

And the error

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/sw/auto/rocky8c/epyc4/software/QIIME2/2022.8/lib/python3.8/site-packages/q2cli/commands.py", line 339, in __call__
    results = action(**arguments)
  File "<decorator-gen-402>", line 2, in denoise_ccs
[...]

PR #168, opened 2024-08-05
IMP: make reverse primer (adapter) optional for denoise-ccs

colinvwood:
addresses #163 

Note that the forward primer is still required because the wrapped dada2 function requires it. This is probably because it is needed to orient the sequences. For users who have neither the forward nor the reverse primer and pre-oriented sequences a deeper solution will be necessary, maybe one that skips the call to `removePrimers` altogether.


lizgehret (approved):
this all looks reasonable to me, thanks @colinvwood - i'll go ahead and merge this. if the forward primer scenario comes up in the future, we can just plan to address separately.

PR #190, opened 2026-03-30
Linked sequences

colinvwood:
# Description
Allow `denoise-paired` to rescue and return unmerged sequences

# AI Disclosure
- [ ] NO AI USED.
- [x] AI USED.

# AI Usage Details
Co-written with Codex


colinvwood:
~~needs qiime2/q2-types#388~~

colinvwood:
@ebolyen

colinvwood:
disclaimer: the `run_dada.R` diff is 99% codex and the `test_denoise.py` diff is about 50/50 codex/de moi

colinvwood:
One outstanding decision is how we want to restructure the stats table to accommodate these changes.

q2_dada2/assets/run_dada.R
```diff
@@ -560,15 +603,42 @@ if(inp.dirR =='NULL'){#for CCS/sinlge/pyro read analysis
 
 }
 
+# combine the merged and unmerged/concatenated feature tables
+if(inp.dirR !='NULL' && retain.unmerged){
```
colinvwood, line 607: do we need inp.dirR guard here

q2_dada2/assets/run_dada.R
```diff
@@ -560,15 +603,42 @@ if(inp.dirR =='NULL'){#for CCS/sinlge/pyro read analysis
...
+    unmerged.ids <- character()
+  }
+}else{
+  unmerged.ids <- character()
+}
 
 ### Remove chimeras
 cat("5) Remove chimeras (method = ", chimeraMethod, ")\n", sep="")
-if(chimeraMethod %in% c("pooled", "consensus")) {
+if(chimeraMethod %in% c("pooled", "consensus") && ncol(seqtab) > 0) {
```
colinvwood, line 626: is `ncol(seqtab) > 0` a necessary guard here?

q2_dada2/assets/run_dada.R
```diff
@@ -545,12 +551,49 @@ if(inp.dirR =='NULL'){#for CCS/sinlge/pyro read analysis
...
+        # are recognized if chimera filtering is performed; this separator
+        # is later converted to a single space
+        unmerged[[j]] <- data.frame(
+          sequence=paste(
+            unmerged.forward, unmerged.reverse,
+            sep=linked.concat.delim
+          ),
+          abundance=unmerged.j$abundance,
+          stringsAsFactors=FALSE
+        )
```
colinvwood, line 589: think about whether we can condense at the sample level instead of below separately. Make sure we still have enough information to create e.g. a number unmerged column in the stats table

q2_dada2/assets/run_dada.R
colinvwood, line 607: this entire block will go away if we do sample level table merging

q2_dada2/assets/run_dada.R
```diff
@@ -560,15 +603,42 @@ if(inp.dirR =='NULL'){#for CCS/sinlge/pyro read analysis
...
   seqtab.nochim <- removeBimeraDenovo(seqtab, method=chimeraMethod, minFoldParentOverAbundance=minParentFold, allowOneOff=allowOneOff, multithread=multithread)
 } else { # No chimera removal, copy seqtab to seqtab.nochim
   seqtab.nochim <- seqtab
 }
 
+# after chimera filtering, convert any retained concatenated IDs to the
+# single space-delimited representation
+if(length(unmerged.ids) > 0){
+  unmerged.keep <- intersect(colnames(seqtab.nochim), unmerged.ids)
+  if(length(unmerged.keep) > 0){
```
colinvwood, line 636: We need to make sure we ONLY replace the 'N' * 10 that we introduced and not any 'N' * 10s that may have already been present in either the forward or reverse (credit to @Oddant1 for noticing this)

colinvwood: at the very least we error, but we know what the forward and reverse sequences are, so we can deduce what the artificially introduced 'N' * 10 is and remove only it

Oddant1 (commented):
@colinvwood's comments came from the review. Needed to look it over with him to make sure I knew what the R was doing.

q2_dada2/assets/run_dada.R
colinvwood: @Oddant1 this isn't actually possible because any reads with any number of N's upstream will be removed by dada2. That said, I changed it to guard against this anyway if that were to somehow change in the future.

colinvwood:
related to #129, #93, #189, qiime2/q2-feature-classifier#219, qiime2/q2-feature-classifier#224

colinvwood:
~~Needs qiime2/q2cli#401~~

q2_dada2/plugin_setup.py
```diff
@@ -24,6 +24,12 @@
 
 _POOL_OPT = {'pseudo', 'independent'}
 _CHIM_OPT = {'consensus', 'none'}
+P_retain_unmerged, T_paired_representative_sequences = qiime2.plugin.TypeMap({
+    qiime2.plugin.Bool % qiime2.plugin.Choices(True):
```
ebolyen:
```suggestion
    qiime2.plugin.Choices(True):
```

q2_dada2/plugin_setup.py
```diff
@@ -24,6 +24,12 @@
...
+        FeatureData[LinkedSequence],
+    qiime2.plugin.Bool % qiime2.plugin.Choices(False):
```
ebolyen:
```suggestion
    qiime2.plugin.Choices(False):
```

q2_dada2/plugin_setup.py
```diff
@@ -181,9 +187,10 @@
                 'n_threads': qiime2.plugin.Threads,
                 'n_reads_learn': qiime2.plugin.Int,
                 'hashed_feature_ids': qiime2.plugin.Bool,
-                'retain_all_samples': qiime2.plugin.Bool},
+                'retain_all_samples': qiime2.plugin.Bool,
+                'retain_unmerged': P_retain_unmerged},
```
ebolyen:
```suggestion
                'retain_unmerged': qiime2.plugin.Bool % P_retain_unmerged},
```

ebolyen (changes requested):
Some grouped suggestions.

nbokulich:
hey drive-by comment: I am testing this with some real data locally to see how it compares to single and paired, looking great! I like how the output stats report the number of concatenated seqs that are rescued, this is really useful. Could we also have a sum total of merged non-chimeric + concatenated, for simplicity we could just call this column "output" (and an additional column of output / input). Very cool to finally have this functionality in q2-dada2, thanks @colinvwood !

q2_dada2/_denoise.py
```diff
@@ -182,18 +202,24 @@ def _denoise_helper(biom_fp, track_fp, err_track_fp,
     # reintroduced above!
     if not retain_all_samples:
         table = table.remove_empty(axis="sample", inplace=False)
+
+    def _to_sequence(sequence, metadata):
+        if retain_unmerged:
+            return skbio.Sequence(sequence, metadata=metadata)
```
ebolyen: I am 99% sure this will still work even with the new LinkedDNA class.

q2_dada2/assets/run_dada.R
```diff
@@ -560,15 +621,36 @@ if(inp.dirR =='NULL'){#for CCS/sinlge/pyro read analysis
...
+# after chimera filtering, convert retained concatenated IDs to the single
+# space-delimited representation using the exact joins we introduced above
+if(nrow(unmerged.id.map) > 0){
+  unmerged.id.map <- unique(unmerged.id.map)
+  ambiguous.ids <- unmerged.id.map$temporary[
+    duplicated(unmerged.id.map$temporary) |
+      duplicated(unmerged.id.map$temporary, fromLast=TRUE)
+  ]
+  if(length(ambiguous.ids) > 0){
+    errQuit("Unable to uniquely map retained unmerged sequences from the temporary DADA2-compatible representation to linked sequences.", status=1)
```
ebolyen, line 640: I am quite confused how this would be possible. Can you explain a little more in a comment?

ebolyen (approved):
I think this looks good, but I did get lost in one spot.

q2_dada2/assets/run_dada.R
colinvwood: Say we have two linked sequences like so:
```
>seq1
ACN TG
>seq2
AC NTG
```

The the `unmerged.id.map` becomes (imagine we're using a single `N` separator instead of ten):
```
ACNNTG -> ACN TG
ACNNTG -> AC NTG
```

Now it's ambiguous what `ACNNTG` maps to.

q2_dada2/_denoise.py
colinvwood: You mean the `__init__` signature?

colinvwood:
Needs qiime2/q2-types#397. (Need to merge and use those changes).

colinvwood:
> hey drive-by comment: I am testing this with some real data locally to see how it compares to single and paired, looking great! I like how the output stats report the number of concatenated seqs that are rescued, this is really useful. Could we also have a sum total of merged non-chimeric + concatenated, for simplicity we could just call this column "output" (and an additional column of output / input). Very cool to finally have this functionality in q2-dada2, thanks @colinvwood !

Thank you for the feedback, I agree the stats table needs polishing. Currently, your "output" column is actually just the "non-chimeric" column. (This is confusing and needs to change.)

Do you think it's useful to see chimera filtering stats stratified by merged/concatenated? If not we could just drop the last two columns (non-chimeric concatenated & its %). That would leave us with only **two** new columns compared to the pre-concatenation changes.

If we do want to see the stratified output we need to add a merged-only & its % columns for chimera filtering. This would leave us with **six** new columns compared to the pre-concatenation version, which feels like it's starting to make the stats table pretty dense. But I can also see the extra stats being useful.

What do you all think @ebolyen, @cherman2?

nbokulich:
> Do you think it's useful to see chimera filtering stats stratified by merged/concatenated? If not we could just drop the last two columns (non-chimeric concatenated & its %).

I don't think we need chimera filtering stats on the concatenated and merged reads separately... could just lump these together into one. I agree, otherwise this table becomes very dense.

Issue #186, opened 2025-08-26
dada2 version; learnErrors() option nreads; q2-dada2 option --p-n-reads-learn


Phylloxera:
What version of dada2 is being used right now? I saw [here](https://github.com/qiime2/q2-dada2/issues/133) that it is >1.16. However, nreads has been deprecated and replaced by nbases since [version 1.7.8](https://github.com/benjjneb/dada2/blob/ab844341c47d8fe922c2e9dea849bf2167bcd445/NEWS#L102), but --p-n-reads-learn is still listed as an option under `qiime dada2 denoise-paired --help` in qiime2-amplicon-2025.7. Can I get some clarification on this apparent version/options mismatch?

Phylloxera:
I confirmed in q2-dada2/assets/run_dada.R that nreads is indeed being applied. and it looks like it is effectively replacing the nbases option inside of dada2::learnErrors since `if (!is.null(nreads) && NREADS > nreads) {
      break
    }` This should mean that the qiime2 default of nreads = 1e6 is more stringent (and time consuming) than the learnErrors default of nbases = 1e8 when the average read length is > 100 and less stringent when the average read length is < 100. Maybe @benjjneb switched to nbases because it removes the effect of read length on the error calculation?

benjjneb:
> Maybe @benjjneb switched to nbases because it removes the effect of read length on the error calculation?

Correct.

lizgehret:
Hey @benjjneb, if our understanding is correct it looks like nreads is still a usable param (we call it [here](https://github.com/qiime2/q2-dada2/blob/b79a17e6d50ad4ec2edfd5f734c6c9d5ff046cb4/q2_dada2/assets/run_dada.R#L452) in q2-dada2) even though it's been deprecated. We are using bioconductor-dada2=1.30.0 which is well past the version where nreads was deprecated. Would you recommend we update our usage of nreads to nbases? If so, when do you expect to fully remove the functionality for nreads?

benjjneb:
> If so, when do you expect to fully remove the functionality for nreads?

I don't intend to remove it.

> Would you recommend we update our usage of nreads to nbases?

I think it's a low priority. The main use case is for long-read sequencing, which is already broken out into a separate q2 action. Leaving nreads as is for the `denoise-single` and `denoise-paired` commands intended for short-read data is fine.

Phylloxera:
[...]

PR #191, opened 2026-04-25
DOC: clarify n_reads_learn help text

Abdeltoto:
## Summary
- Clarify that `n_reads_learn` is passed through as `--learn_min_reads`.
- Apply the same help text clarification across the denoise actions that expose `n_reads_learn`.

## Test plan
- `python -m py_compile q2_dada2/plugin_setup.py`

## AI disclosure
AI assistance was used for codebase navigation and wording suggestions. I reviewed the relevant q2-dada2 code paths and edited the final change before submitting.

Refs #186.

### DIFF
--- a/q2_dada2/plugin_setup.py
+++ b/q2_dada2/plugin_setup.py
@@ -123,9 +123,10 @@ plugin.methods.register_function(
                       'processing. If 0 is provided, all available cores will '
                       'be used.'),
         'n_reads_learn': ('The number of reads to use when training the '
-                          'error model. Smaller numbers will result in a '
-                          'shorter run time but a less reliable error '
-                          'model.'),
+                          'error model. This is passed to DADA2 as the '
+                          '`nreads` argument to `learnErrors`. Smaller '
+                          'numbers will result in a shorter run time but a '
+                          'less reliable error model.'),
         'hashed_feature_ids': ('If true, the feature ids in the resulting '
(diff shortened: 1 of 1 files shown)


gregcaporaso:
@Abdeltoto, this help text isn't correct - the value gets passed as `--learn_min_reads` to DADA2. If you'd like to update that, we'll accept this PR (though note that we are planning to replace this parameter so this will be a short-lived addition). 

Please also review our [AI Code Policy](https://github.com/rachis-org/governance/blob/main/ai_policy.md) and ensure that your disclosures are in place as needed.

Abdeltoto:
Thanks for the correction. I updated the help text to describe the value as being passed through as \--learn_min_reads\ instead of \
reads\, and added an AI disclosure to the PR description after reviewing the linked policy.

Local check: \python -m py_compile q2_dada2/plugin_setup.py\.

gregcaporaso (approved):
Looks good now, thanks for the addition @Abdeltoto.

colinvwood:
I don't think this is quite right. The way @Abdeltoto had it the first time is correct. The `--learn-min-reads` parameter is only something our internal `run_dada.R` script recognizes; it goes on to be passed to `nreads` in `learnErrors` which is actually part of dada2's API.

gregcaporaso:
Ah, sorry if I got that wrong. If that's the case @colinvwood, could you coordinate a fix on this?

Issue #172, opened 2024-12-03
IMP: restructure dada2 invocation from R script to rpy2


colinvwood:
Currently we use an R script (`q2_dada2/assets/run_dada.R`) to run the dada2 pipeline for all four denoise methods in this plugin. Thus separate logic for each of these actions exists in each of the python functions and is also intermingled in the R script. Instead we should use the `rpy2` library to interact with dada2's API within each method, increasing modularity. This will also make the plugin more maintainable because our team is more familiar with python than with R.

colinvwood:
~~This should probably wait until after #169 is merged.~~

gregcaporaso:
~~This may end up causing issues since DADA2 is GPL-licensed. We should keep this in mind.~~

colinbrislawn:
>This will also make the plugin more maintainable because our team is more familiar with python than with R.

I'm not an expert like Ben, but I'm happy to help clean up the R side of this codebase.

colinvwood:
Hey @colinbrislawn the idea behind this issue is to move the R logic into python, not refactor the R... That said, we haven't made a definitive decision about whether we actually want to implement this, so in the future cleaning up the R script may be welcome (doing it now could be a waste of time if we end up using rpy2).

PR #197, opened 2026-07-08
Refactors `_run_dada` from R to python

Macabe222:
# Description

Deletes `run_dada.R` and replaces it with `_run_dada.py`, using  rpy2. Slightly faster on moving pictures demux.qza, 1:28 vs 1:33.
Addresses #172 
# AI Disclosure

- [X] NO AI USED.
- [ ] AI USED.


colinvwood:
~~waiting on #190~~

colinvwood:
Note: Pseudo-pooling now uses DADA2’s native pool="pseudo" implementation, instead of two calls to `dada` on our end. That updates the error model between rounds, whereas run_dada.R reused the first learned model. This results in different, though desirable, output compared to the pre-refactor version.

benjjneb (changes requested):
The dada2 derepFastq function should not be called explicitly in this workflow. This is to avoid loading all samples into memory at the same time unnecessarily, which can be performance-breaking behavior on larger datasets.

https://github.com/benjjneb/dada2/issues/919#issuecomment-573907802
https://github.com/benjjneb/dada2/issues/1095#issuecomment-671333174

The R `dada` function should be passed a vector of file paths rather than the outpur of `derepFastq`.

Specific places where this performance hit will be felt are in the pseudo-pooling calls:

https://github.com/qiime2/q2-dada2/pull/197/changes/d638c023f9ab03f9a17061633b559ed7a37f63a1..0fc08b57a3e40a53b741dc459ee163cb6101c222#diff-dd7407f67ac9fa623337b0138c3539d679aa730b1ab132271bbf21a218822fbeR439
https://github.com/qiime2/q2-dada2/pull/197/changes/d638c023f9ab03f9a17061633b559ed7a37f63a1..0fc08b57a3e40a53b741dc459ee163cb6101c222#diff-dd7407f67ac9fa623337b0138c3539d679aa730b1ab132271bbf21a218822fbeR545
https://github.com/qiime2/q2-dada2/pull/197/changes/d638c023f9ab03f9a17061633b559ed7a37f63a1..0fc08b57a3e40a53b741dc459ee163cb6101c222#diff-dd7407f67ac9fa623337b0138c3539d679aa730b1ab132271bbf21a218822fbeR552

But the code for the regular operation of `dada` can also be simplified by removing the looping in the Q2 code:

https://github.com/qiime2/q2-dada2/pull/197/changes/d638c023f9ab03f9a17061633b559ed7a37f63a1..0fc08b57a3e40a53b741dc459ee163cb6101c222#diff-dd7407f67ac9fa623337b0138c3539d679aa730b1ab132271bbf21a218822fbeR568

Similarly, the calls to `mergePairs` can also take the filepaths instead of the output of `derepFastq`.

A test (probably non-automatable due to computational cost) to see if this is working correctly is to evaluate the memory usage of this workflow on increasing sample number. At larger numbers of samples, memory costs if `derepFastq` is being called on the whole dataset will go up linearly with sample number, while if samples are being processed one-at-a-time the memory cost will stay roughly constant with sample number.

colinvwood:
Hello @benjjneb, thank you for the feedback. The memory concerns make sense; I will refactor to pass filepaths and let the R package be more intelligent wrt the derep objects.

q2_dada2/tests/test_denoise.py
```diff
@@ -34,13 +39,86 @@ def _sort_table(table):
     return table.sort(axis="sample").sort(axis="observation")
 
 
+def _assert_error_models_equal(actual, expected):
+    assert_frame_equal(
+        actual.to_dataframe().replace('', pd.NA),
+        expected.to_dataframe().replace('', pd.NA)
+    )
```
Oddant1, line 46: Very good

q2_dada2/_r_utils.py
```diff
@@ -0,0 +1,40 @@
...
+# ----------------------------------------------------------------------------
+
+import pandas as pd
+
+from rpy2.rinterface import NULL
+from rpy2.robjects import conversion, default_converter, pandas2ri, RObject
+from rpy2.robjects.vectors import DataFrame as RDataFrame, Matrix as RMatrix
+
+
+def _robj_to_pandas_df(robj: RObject | pd.DataFrame) -> pd.DataFrame:
```
Oddant1, line 16: Can't rpy2 do this natively? Why do we need this?

conda-recipe/meta.yaml
```diff
@@ -24,6 +24,7 @@ requirements:
   - openjdk
   - qiime2 >={{ qiime2 }}
   - q2-types >={{ q2_types }}
+  - rpy2 {{ rpy2 }}
```
Oddant1, line 27: 👎

q2_dada2/_dada_stats/__init__.py
```diff
@@ -6,7 +6,12 @@
...
+__all__ = ['plot_base_transitions']
 
 
-__all__ = ['plot_base_transitions']
+def __getattr__(name):
+    if name == 'plot_base_transitions':
+        from ._visualizer import plot_base_transitions
+        return plot_base_transitions
+
+    raise AttributeError(f'module {__name__!r} has no attribute {name!r}')
```
Oddant1: What on Earth are these shenanigans

Oddant1 (commented):
Preliminary review. More to come.

One additional thing, what's with the changes to the test data?

colinvwood:
> Preliminary review. More to come.
> 
> One additional thing, what's with the changes to the test data?

Good question, they're for the pseudo pooling method only, and the reason is explained in my comment from a few weeks ago:

> Note: Pseudo-pooling now uses DADA2’s native pool="pseudo" implementation, instead of two calls to dada on our end. That updates the error model between rounds, whereas run_dada.R reused the first learned model. This results in different, though desirable, output compared to the pre-refactor version.

q2_dada2/_r_utils.py
colinvwood: to restore the row names and colnames I believe which doesn't happen by default, I'm assuming (Macabe wrote the original version of this)

Oddant1: The names are maintained for DataFrames. I suppose the comment does also say it does R matrices -> pd.DataFrame and idk what that entails for rpy2. @Macabe222 can you shed some light on this?

EDIT: Looking back on it, it looks like if it is a pd.DataFrame it just returns it anyway or uses the built in conversion and returns it... I guess this does keep the conversion in one place. Idk I suppose I'm fine with it. Just didn't look too too close yesterday.

q2_dada2/_dada_stats/_error_model.py
```diff
@@ -0,0 +1,138 @@
...
+        trans_df = _melt_error_matrix(trans)
+        trans_df.columns = ['Transition', 'Qual', 'count']
+    elif err_out is not None:
+        if len(err_out.columns) <= 1:
+            raise ValueError(
+                'plotErrors only supported when using quality scores in the '
+                'error model (i.e. USE_QUALS=TRUE).'
+            )
+        trans_df = _melt_error_matrix(err_out)
+        trans_df.columns = ['Transition', 'Qual', 'count']
```
Oddant1: I don't think this if/elif reads well. It isn't clear why we'd only have one or the other or what the difference is between having one vs the other.

colinvwood: Seems like we can tighten our assumptions here and expect `$trans` to always be present.

lizgehret:
Just chatted with @ebolyen about this, we'll get this reviewed shortly!

Issue #188, opened 2025-10-28
Expose Ability to Explicitly Set Phred-Offset in DADA2


Oddant1:
**Improvement Description**
Expose the ability to manually tell DADA2 the phred-offset of our data.

**Current Behavior**
As far as I can tell, we do not currently offer a way to explicitly tell DADA2 the phred-offset of the data we are passing into it. This causes issues when DADA2 automatically determines the phred-offset incorrectly (see forum x-ref).

**Proposed Behavior**
We should be able to tell DADA2 our phred-offset

**Questions**
1. How feasible is this? Based on the forum x-ref, it seems like there isn't just one parameter that controls this.

**References**
1. https://forum.qiime2.org/t/dada2-fails-on-aviti-data/33741

Oddant1:
Based on updates to the forum x-ref, this may be handled internally to DADA2, and then we just need to update DADA2

PR #199, opened 2026-08-24
Expose `qualityType` parameter to fix DADA2 misdetection of AVITI/high-quality FASTQ encoding

vaamb:
While processing AVITI sequencing data (excellent quality, Q40–Q45), the denoising step consistently failed with:
`Invalid derep$quals matrix. Quality values must be positive integers`

This traces back to `benjjneb/dada2#1549`.
Internally, dada2 relies on ShortRead to load FASTQ files, and ShortRead auto-detects the quality encoding as either `FastqQuality` (ASCII 33–126, phred-offset 33) or `SFastqQuality` (ASCII 64–126, phred-offset 64). Under certain conditions, AVITI reads are misidentified as `SFastqQuality`, causing an offset of 64 to be applied instead of 33 and producing the error above

This patch exposes the `qualityType` parameter from `ShortRead` / `dada2` to `q2-dada2`, allowing users to explicitly set the encoding and bypass the faulty auto-detection

A (cleaner) fix would be for q2-dada2 to set `qualityType="FastqQuality"` by default, since it seems QIIME2 already enforces phred-offset 33 upstream

AI Disclosure
Not used

### DIFF
--- a/q2_dada2/_denoise.py
+++ b/q2_dada2/_denoise.py
@@ -221,7 +221,7 @@ def _denoise_single(demultiplexed_seqs, trunc_len, trim_left, max_ee, trunc_q,
                     max_len, pooling_method, chimera_method,
                     min_fold_parent_over_abundance, allow_one_off,
                     n_threads, n_reads_learn, hashed_feature_ids,
-                    homopolymer_gap_penalty, band_size, retain_all_samples):
+                    homopolymer_gap_penalty, band_size, retain_all_samples, quality_type):
     _check_inputs(**locals())
     if trunc_len != 0 and trim_left >= trunc_len:
         raise ValueError("trim_left (%r) must be smaller than trunc_len (%r)"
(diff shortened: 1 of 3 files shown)


vaamb:
This is a fix for qiime2/q2-dada2#188

vaamb:
For the patch of the custom version we use in my lab I finished using the `derepFastq({...}, qualityType='FastqQuality')` as it is invisible to the end user, and solves all the issues encountered with the AVITI sequences

colinvwood:
Hello @vaamb, thank you for the PR. Unfortunately, we can't merge as-is due to an extensive refactor underway in #197. However, once we have that merged I can port your changes and will give you credit for them. I will investigate the simpler solution you mentioned (always assume phred-33) at that time. Does that sound alright?

vaamb:
Hi @colinvwood !

The simpler solution is cleaner IMO. I'll force push the changes here so this pull request will highlight just what is needed.
I also ended up to sprinkle `qualityType=FastqQuality` in some other functions to improve the quantity of reads (rightly) passing the initial filtering step. But the plugin now works with AVITI sequences !

PR #200, opened 2026-09-23
SKIP: Remove openjdk dependency from meta.yaml

ebolyen:
# Description

There used to be a comment explaining this: https://github.com/qiime2/q2-dada2/blob/b47f488f60d77f3650f3f9e348ec03db8d94b044/ci/recipe/meta.yaml#L25-L32

Based on the comment, I tested `java` and it responds on stdout now instead of creating the GUI popup, so I think this is no longer an issue.

# AI Disclosure

- [ ] NO AI USED.
- [x] AI USED.

# AI Usage Details

I was researching our transitive dependencies for a different purpose and that tooling involved AI.


ebolyen:
Tested in a fresh environment with R: `R CMD javareconf` no longer creates the popup and functions as one would expect.

